PR #14, opened 2019-03-25
Preliminary UI Integration

nwingt:


src/components/PageHeader.vue
```diff
@@ -0,0 +1,151 @@
...
+        transform: `translateY(${this.y}px)`,
+      };
+      if (this.scrollY === 0) {
+        style.visibility = 'hidden';
+      }
+      return style;
+    },
+  },
+  mounted() {
+    this.throttledUpdateLayout = throttle(this.updateLayout, 16);
```
williamchong007: should use `window.requestAnimationFrame()` instead

williamchong007:
Please remove vuetify from `functions/package(-lock).json` too

src/util/api/index.js
```diff
@@ -17,3 +17,15 @@ export const getOAuthLoginAPI = () => '/api/users/login';
...
 export const getLoginStatus = () => `/api/users/self`;
+
+export const getUserMinAPI = likerId =>
+  `https://${
+    IS_TESTNET ? 'api.rinkeby.like.co' : 'api.like.co'
+  }/api/users/id/${likerId}/min`;
+
+export const getLikeButtonTotalLikeCountAPI = (likerId, referrerURL) =>
+  `https://${
+    IS_TESTNET ? 'rinkeby.like.co' : 'like.co'
```
williamchong007: this api is also accessible with `api.(rinkeby.)like.co`, so maybe make `api.(rinkeby.)like.co` a constant  `LIKE_CO_API_BASE`
also in `api.(rinkeby.).like.co` you can omit the `/api` part

PR #16, opened 2019-03-26
🌐 Add auto i18n detection

williamchong007:


src/plugins/vue-i18n.js
```diff
@@ -1,17 +1,49 @@
...
+      defaultLocale;
+    // TODO: iterate through navigator.languages to find locale
+    navLang = navLang.toLowerCase();
+    availableLocales.forEach(key => {
+      if (navLang.includes(key)) {
+        navLang = key;
+      }
+    });
+    locale =
+      locale ||
```
nnkken: If `defaultLocale` is set, then all works will be wasted?

PR #22, opened 2019-03-27
Add settings page

nwingt:


src/pages/settings.vue
```diff
@@ -26,6 +26,11 @@ export default {
     SiteNavBar,
     SettingsPageHeader,
   },
+  computed: {
+    isHeaderShowBack() {
+      return /settings-.+/.test(this.$route.name);
```
williamchong007: regex should check ^?

src/pages/settings/following/index.vue
```diff
@@ -0,0 +1,98 @@
...
+
+export default {
+  components: {
+    AuthorListItem,
+  },
+  computed: {
+    ...mapGetters(['getSubscribedAuthors', 'getUnsubscribedAuthors']),
+
+    authorLikerIds() {
+      if (this.$route.name === 'settings-following') {
```
williamchong007: `this.$route.name === 'settings-following'`  should be abstracted to a computed property `is(Un)subscribeSetting`

PR #27, opened 2019-03-28
Update civic page UI and fix home page CTA

nwingt:


functions/package.json
```diff
@@ -33,6 +33,7 @@
     "portal-vue": "^1.5.1",
     "source-map-support": "^0.5.10",
     "vue": "^2.6.7",
+    "vue-awesome-swiper": "^3.1.3",
```
williamchong007: non-ssr plugin is not needed in `functions/package.json`

PR #32, opened 2019-04-01
⚡️ Lazyload og images

williamchong007:


src/nuxt.config.js
```diff
@@ -51,8 +51,8 @@ module.exports = {
   plugins: [
     '~/plugins/likecoin-ui-vue.js',
     '~/plugins/portal-vue.js',
-    { src: '~/plugins/vue-awesome-swiper.js', ssr: false },
```
nnkken: Why this is deleted?

williamchong007: moved to `client-ui-plugins.js`

PR #40, opened 2019-04-02
🎨 Rename client plugin to follow nuxt plugin name conventional

williamchong007:
✨ Attempt to use IntersectionObserver in vue-lazy


nwingt:
What is the difference between this and #39

williamchong007:
@nwingt push wrong, will force push later

williamchong007:
@nwingt please review again

PR #41, opened 2019-04-02
⚡️ Reduce api payload size by sending url on-demand

williamchong007:


williamchong007:
@nnkken should I send `referrer` on demand or `url` on demand...

PR #52, opened 2019-04-04
 ✨ Add image resize cloud function

williamchong007:


src/util/api/index.js
```diff
@@ -28,6 +28,12 @@ export const getCivicLikerTrialEventByIdAPI = id =>
 export const getCivicLikerJoinTrialEventByIdAPI = id =>
   `/api/civic/trial/events/${id}/join`;
 
+export const getImageResizeAPI = (url, { width } = {}) =>
+  `https://us-central1-civic-liker${
+    IS_TESTNET ? '-develop' : ''
+  }.cloudfunctions.net/thumbnail/?url=${encodeURIComponent(url)}${
```
nwingt: Should pass host in env var?

williamchong007: probably want to use it in dev env too to avoid unwanted diff in behaviour...
not sure

nwingt: can others use that API publicly? If so, is that a concern?

packages/resizeImageHttp/index.js
```diff
@@ -0,0 +1,57 @@
...
+  IS_TESTNET,
+} = process.env;
+
+const app = express();
+
+app.use(helmet());
+app.get('/', async (req, res) => {
+  const { url } = req.query;
+  const { origin } = req.headers;
+  if (!IS_TESTNET && origin && !origin.includes(ORIGIN_DOMAIN)) {
```
nnkken: What if the origin header is not in the request?
Should use CORS here to enforce?

williamchong007: browser does not enforce cors on img src

nnkken: So the checking is only for best effort?

src/util/api/index.js
williamchong007: probably can as in we didnt put any real checking here except origin header, and this url is very visible under browser network inspector

packages/resizeImageHttp/index.js
williamchong007: yes unless we check against db record of og image

PR #53, opened 2019-04-08
Fixes before launch

nwingt:
Better view with ignore white space option on and commit by commit

### DIFF
--- a/src/mixins/sliding-menu.js
+++ b/src/mixins/sliding-menu.js
@@ -17,26 +17,26 @@ export default {
     };
   },
   beforeDestroy() {
-    this.toggleClickListener(false);
+    this.manageClickListener(false);
   },
   watch: {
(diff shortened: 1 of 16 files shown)


src/mixins/sliding-menu.js
```diff
@@ -27,13 +27,17 @@ export default {
...
-    toggleClickListener(isOn) {
+    toggleEventListener(eventName, isOn) {
       window[`${isOn ? 'add' : 'remove'}EventListener`](
-        'click',
+        eventName,
         this.onWindowClick,
         true
       );
     },
+    toggleClickListener(isOn) {
```
williamchong007: not sure about `toggle` `isOn` is make it on or toggle it off from on
just use sth like `isAdd` / `isRemove`?

nwingt: Updated

PR #54, opened 2019-04-08
Show bookmark button even if user hasn't signed in

nwingt:


williamchong007:
@nwingt can you plz help apply similar logic onto the register button in the civic liker page? thanks

PR #58, opened 2019-04-08
🎨 Rewrite referrer checking logic in resize image api

williamchong007:


packages/resizeImageHttp/index.js
```diff
@@ -15,14 +16,20 @@ const app = express();
...
-  const { origin, referer } = req.headers;
-  if (!IS_TESTNET && origin && !origin.includes(ORIGIN_DOMAIN)) {
-    res.sendStatus(403);
-    return;
-  }
-  if (!IS_TESTNET && referer && !referer.includes(ORIGIN_DOMAIN)) {
-    res.sendStatus(403);
-    return;
+  const { referer } = req.headers;
+  if (referer) {
```
nnkken: Why drops the `origin` header checking?

williamchong007: at least chrome does not send it when fetching img src, only referrer is sent
.... maybe useful for a canvas loaded img though, not sure

nnkken: But then keeping it should not harm, and some browser does not send `referer` header, so keeping the `origin` should in theory catch more cases?

williamchong007: i doubt if they dont send `referer` would they send `origin`....
still we can keep it

PR #64, opened 2019-04-10
🚸 Show bookmark from latest to oldest

williamchong007:


src/pages/index/bookmarks.vue
```diff
@@ -58,7 +58,8 @@ export default {
   computed: {
     ...mapGetters(['getUserBookmarks']),
     items() {
-      return this.getUserBookmarks.map(referrer => ({ referrer }));
+      const items = this.getUserBookmarks.map(referrer => ({ referrer }));
+      return items.reverse();
```
nwingt: Not reverse in API?

williamchong007: seems presentation requirement, so do in UI for now

PR #91, opened 2019-04-25
Move Civic Liker Trial page from like.co

nwingt:


williamchong007:
should thos a link redirect to home be nuxt link instead?
Can save some loading time

nwingt:
> should thos a link redirect to home be nuxt link instead?
> Can save some loading time

Updated

PR #94, opened 2019-04-26
🎨 Split sign in and sign up url

williamchong007:


src/components/SlidingMenu.vue
```diff
@@ -106,7 +106,7 @@ export default {
     ]),
   },
   methods: {
-    getOAuthLoginAPI,
+    getOAuthRegisterAPI,
```
nwingt: use computed instead?

src/server/api/routes/users.js
```diff
@@ -33,10 +33,16 @@ router.get('/users/self', async (req, res, next) => {
   }
 });
 
+router.get('/users/register', (req, res) => {
+  const state = crypto.randomBytes(8).toString('hex');
+  req.session.state = state;
+  res.redirect(getOAuthURL({ state, isLogin: false }));
+});
+
```
nnkken: How about this?

```JS
const loginOrRegisterRoute = ({ isLogin }) => (req, res) => {
  const state = crypto.randomBytes(8).toString('hex');
  req.session.state = state;
  res.redirect(getOAuthURL({ state, isLogin }));
};

router.get('/users/register', loginOrRegisterRoute({ isLogin: false }));
router.get('/users/login', loginOrRegisterRoute({ isLogin: true }));
```

nwingt: How about this
```javascript
function newSessionState(req, res, next) {
  req.session.state = crypto.randomBytes(8).toString('hex');
  next()
}

...

router.get('/users/register', newSessionState, (req, res) => {
  res.redirect(getOAuthURL({ state: req.session.state, isLogin: false }));
});

router.get('/users/login', newSessionState, (req, res) => {
  res.redirect(getOAuthURL({ state: req.session.state, isLogin: true }));
});
```

williamchong007:
```
function setSessionOAuthState(req) {
  const state = crypto.randomBytes(8).toString('hex');
  req.session.state = state;
}



router.get('/users/register', (req, res) => {
  setSessionOAuthState(req);
  res.redirect(getOAuthURL({ state: req.session.state, isLogin: false }));
});

```
?

nwingt: I mean using middleware

PR #95, opened 2019-04-29
⚡️ Add img/api srcs into workbox caching

williamchong007:
mostly work in offline mode for now

### DIFF
--- a/src/nuxt.config.js
+++ b/src/nuxt.config.js
@@ -4,11 +4,17 @@ const theme = require('./tailwind.config');
 
 const siteName = 'Liker.Land';
 
+const {
+  IS_TESTNET,
+  CI,
+  INTERCOM_APPID,
+} = process.env;
+
 const nuxtConfig = {
   env: {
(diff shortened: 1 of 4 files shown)


src/nuxt.config.js
```diff
@@ -162,6 +168,16 @@ const nuxtConfig = {
...
    * Nuxt PWA modules
    */
+  workbox: {
+    runtimeCaching: [
+      {
+        urlPattern: `https://api\\.${IS_TESTNET ? 'rinkeby\\.' : '' }like\\.co/.*`,
+      },
+      {
+        urlPattern: `https://us-central1-civic-liker${IS_TESTNET ? '-develop' : '' }\\.cloudfunctions\\.net/.*`,
+      },
```
nwingt: Not regex?

nnkken: According to [doc](https://pwa.nuxtjs.org/modules/workbox.html#adding-custom-runtimecaching-items-for-cdn), the string will be compiled by `new RegExp(...)`

PR #97, opened 2019-04-30
Update civic page using starter pack content

nwingt:


src/components/LikerComparisonCard.vue
```diff
@@ -85,6 +54,48 @@ export default {
...
+    },
+  },
+  methods: {
+    getFeatureAvailability(feature) {
+      return this.plan.features.includes(feature);
+    },
+    getFeatureText(feature) {
+      const i18nKey = `LikerFeature.${feature}`;
+      if (this.$t(i18nKey, this.$i18n.fallbackLocale) instanceof Object) {
+        return this.$t(`${i18nKey}.${this.type}`);
```
williamchong007: should use `$te()` to check for key exist

nwingt: Updated

PR #107, opened 2019-05-08
🌐 Add cantonese video in civic

williamchong007:


src/pages/civic/index.vue
```diff
@@ -145,14 +145,27 @@ export default {
...
       }
     },
+    isCantonese() {
+      if (!process.client) return false;
+      return (
+        navigator &&
+        ((navigator.language &&
+          navigator.language.toLowerCase().includes('hk')) ||
+          (navigator.languages &&
+            navigator.languages.find(lang =>
```
nwingt: suppose navigator.language is one of navigatior.languages?

williamchong007: ` In Chrome, navigator.language is the language of the browser UI, and is not guaranteed to be equal to navigator.languages[0].`
not sure if thats good or bad..

PR #111, opened 2019-05-09
⚡️ Skip store init if no cookie exist

williamchong007:
⚡️ Clear cookie when logout

### DIFF
--- a/src/store/index.js
+++ b/src/store/index.js
@@ -9,14 +9,17 @@ import reader from './modules/reader';
 import ui from './modules/ui';
 import user from './modules/user';
 import staticData from './modules/staticData';
+import { AUTH_COOKIE_NAME } from '~/constant';
 
 const createStore = () =>
   new Vuex.Store({
     actions: {
-      async nuxtServerInit({ commit }, ctx) {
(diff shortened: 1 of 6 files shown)


src/store/index.js
```diff
@@ -13,10 +13,13 @@ import staticData from './modules/staticData';
 const createStore = () =>
   new Vuex.Store({
     actions: {
-      async nuxtServerInit({ commit }, ctx) {
+      async nuxtServerInit({ commit }, { req }) {
         try {
-          const userInfo = await this.$axios.$get(api.getLoginStatus());
-          commit(types.USER_SET_USER_INFO, userInfo);
+          // eslint-disable-next-line no-underscore-dangle
+          if (req.cookies && req.cookies.__session) {
```
nnkken: Should access by the `AUTH_COOKIE_NAME` constant?

nwingt: skip one comment line too

williamchong007: this is in frontend, `AUTH_COOKIE_NAME` was defined in `/server
dont really want to import from server constant, any suggestions?
another `constant` in frontend？

PR #126, opened 2019-05-27
Add billing summary

nwingt:

PR #130, opened 2019-06-04
Minor adjustment on sliding menu & settings page

nwingt:


src/components/SlidingMenu.vue
```diff
@@ -104,8 +104,20 @@ export default {
...
       'getUserCivicLikerHalo',
+      'getUserIsCivicLiker',
+      'getUserIsCivicLikerTrial',
       'getHomeRoute',
     ]),
+
+    menuItemTitleForCivic() {
+      if (
+        (this.getUserId && !this.getUserIsCivicLiker) ||
+        this.getUserIsCivicLikerTrial
```
williamchong007: `this.getUserId && (!this.getUserIsCivicLiker || this.getUserIsCivicLikerTrial)` ?

nwingt: The condition is Liker OR Civic Liker (Trial)

williamchong007: but one cannot be a `Civic Liker (Trial)` without `getUserId()` being true?

nwingt: How about I wait for #126 to merge and I will use the new getter...

williamchong007: port the getter from #126 instead? this PR seems simpler and should merge first

nwingt: Ported

PR #131, opened 2019-06-12
Add WelcomeDialog

nwingt:
ezgif-2-e6f066135dae

### DIFF
--- a/src/server/api/routes/users.js
+++ b/src/server/api/routes/users.js
@@ -35,10 +35,10 @@ router.get('/users/self', async (req, res, next) => {
         apiFetchUserProfile(req),
         userCollection.doc(user).get(),
       ]);
-      const { hasReadFirstTimeSignInDialog } = userDoc.data();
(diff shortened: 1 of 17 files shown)


src/server/api/routes/users.js
```diff
@@ -46,6 +55,25 @@ router.get('/users/self', async (req, res, next) => {
...
+      await userCollection.doc(user).update({
+        hasReadFirstTimeSignInDialog,
+      });
+      res.sendStatus(200);
+      return;
+    }
+    res.sendStatus(404);
+  } catch (err) {
+    if (req.session) req.session = null;
+    res.clearCookie(AUTH_COOKIE_NAME, CLEAR_AUTH_COOKIE_OPTION);
```
williamchong007: `/users/self` was decided to be called every time user reach our site, so it also handle invalid session
for other (authed) API endpoint, just throw 403 or slient ignore if session is not right, no need to handle

PR #158, opened 2019-07-31
⚗️ Adjust buttons' order in error page

nwingt:


src/experiments/civic/index.js
```diff
@@ -1,6 +1,6 @@
 export default {
   name: 'civic-register-page-error',
   experimentID: '-ZAD0MyxTySnOmA902g6Jg',
-  isEligible: ({ route }) => !process.server && route.name === 'civic-register',
+  isEligible: () => !process.server,
```
williamchong007: if only want to test ppl directly landed on `/civic`, can keep the route check with name `civic`

nwingt: reverted

williamchong007: `civic` not `civic-register`

williamchong007: this `isEligible` hook only trigger first time ppl arrive the site, so its kind of strange

nwingt: updated

PR #177, opened 2019-10-21
🌐 Add `isHK` geoip detection for payment method list

williamchong007:


src/store/modules/ui.js
```diff
@@ -3,20 +3,28 @@
...
+  UI_SET_LOCALE,
+  UI_SET_IS_HK,
+} from '../mutation-types';
 
 import { defaultLocale, availableLocales } from '../../locales';
 
 const initialState = () => ({
   locales: availableLocales,
   locale: defaultLocale,
+  isHK: true,
```
nwingt: how about storing location with getter isHK?

williamchong007: but I have country and city judgement, in that case need to store 2 state

nwingt: i foresee there will be isTW isCN isJP... etc 
nonetheless this is ok for now

PR #197, opened 2019-11-19
⚗️ Start new exp for /civic with referrer logo

nwingt:


src/pages/civic/index.vue
```diff
@@ -147,8 +142,12 @@
                 .flex.justify-center.mb-24(
                   v-if="isExperimenting"
                 )
-                  LikeButtonAnimation(:size="180")
+                  LikeButtonAnimation(
+                    :size="180"
+                    :avatar="referrer.avatar"
```
williamchong007: will this explode if referrer is undefined?

src/pages/civic/index.vue
```diff
@@ -147,8 +142,12 @@
...
+                  )
                   LcChopCivicLiker(
+                    v-if="!referrer.avatar"
```
williamchong007: will this explode if referrer is undefined?

src/pages/civic/index.vue
```diff
@@ -237,7 +236,12 @@ export default {
   },
   mixins: [
     IntercomMixinFactory({ isBootAtMounted: false }),
-    experimentMixin('isExperimenting', 'civic-page', 'variant'),
+    experimentMixin(
+      'isExperimenting',
+      'civic-page',
+      'variant',
+      that => !!that.referrer
```
nwingt: I have guarded in L243

src/pages/civic/index.vue
nwingt: I have guarded in L243

src/pages/civic/index.vue
nwingt: I have guarded in L143 & L243

src/pages/civic/index.vue
williamchong007: After we end the experiment, L243 would be removed, and this part of template might be applied.
Should also guard in template?

PR #208, opened 2020-01-09
🐛 Do not prompt app banner if useragent is LikeCoinApp

williamchong007:


williamchong007:
Please don't merge yet, confirming if likecoin app set client user agent properly

williamchong007:
Should work after https://github.com/likecoin/likecoin-app/pull/91 is released

PR #213, opened 2020-01-13
🐛 Use parse-url instead of URL to support older browsers

williamchong007:


src/components/ContentCard.vue
```diff
@@ -152,6 +152,8 @@ import TickIcon from '~/assets/icons/tick.svg';
 import { checkIsMobileClient } from '~/util/client';
 import { getAvatarHaloTypeFromUser } from '~/util/user';
 
+const parse = require('url-parse');
```
nnkken: Why mixing `import` and `require`?

PR #218, opened 2020-01-16
💩 Workaround set-cookie header issue in firebase hosting

williamchong007:


src/server/api/index.js
```diff
@@ -51,6 +51,26 @@ router.use(
     unset: 'destroy',
   })
 );
+router.use((req, res, next) => {
+  // HACK: mitigate issue that express-session set-cookie Expires contains comma
+  // which makes firebase hosting wrongly split the set-cookie header into 2
+  // mitigate by overide cookie serialize function and use MaxAge instead of Expires
+  Object.defineProperty(req.session.cookie, 'data', {
```
nnkken: This object will be unwritable (writing will not throw error, but takes no effect), is this the wanted behaviour?

williamchong007: I was overriding the original getter with maxage replacing expires
https://github.com/expressjs/session/blob/master/session/cookie.js#L117

PR #220, opened 2020-01-29
🐛 Fix exp eligibility

nwingt:


src/experiments/civic/index.js
```diff
@@ -1,7 +1,6 @@
 export default {
   name: 'civic-register-page',
   experimentID: 'QBxvQ7G9RpSRypN7noIvOw',
-  isEligible: ({ route }) =>
-    !process.server && /^civic-register.*/.test(route.name),
+  isEligible: ({ route }) => !process.server && /^civic.*/.test(route.name),
```
nnkken: Actually you don't need the `.*` suffix

williamchong007:
I think the experiment condition checking does not trigger on client route change, so it will be always false if not server.
Can you confirm exp cookie was set in dev?

PR #224, opened 2020-02-06
⚗️ Flip experiment variant for test

nwingt:
Tested on Rinkeby


williamchong007:
@nwingt I think flipping the `if (isExperimenting)` in html/js is more robust, since it flip default case too?

PR #231, opened 2020-02-23
Fix payment selection for non HK user

DarkSorrow:
Change detection flag as non HK since it's the default behaviour and added browser language detection to have more accuracy

### DIFF
--- a/src/pages/civic/index.vue
+++ b/src/pages/civic/index.vue
@@ -385,13 +385,28 @@ export default {
       ],
     };
   },
-  mounted() {
-    if (this.getIsHK) {
+  async beforeMount() {
+    let isHK = this.getIsHK;
+    if (isHK === undefined) {
+      isHK = false; // Default not from HK
+      try {
+        const { data: geoData } = await this.$axios.get('/api/civic/geoip');
+        if (geoData.ipCountry || geoData.ipCity) {
+          isHK = geoData.ipCountry === 'HK' || geoData.ipCity === 'hong kong';
+        }
+      } catch (err) {
+        isHK = false;
+      }
+      this.$store.dispatch('setIsHK', isHK);
+    }
+    if (isHK) {
       this.selectedPaymentMethod =
         PAYMENT_METHOD_LIST[PAYMENT_METHOD_LIST.length - 1];
       this.$i18n.locale = 'zh-Hant';
       this.setLocale(this.$i18n.locale);
     }
+  },
+  mounted() {
     const {
       from,
       referrer,
--- a/src/pages/civic/index.vue
+++ b/src/pages/civic/index.vue
@@ -214,7 +214,7 @@ import { getAvatarHaloTypeFromUser, checkUserNameValid } from '~/util/user';
 import { IntercomMixinFactory } from '~/mixins/intercom';
 import experimentMixin from '~/mixins/experiment';
 
-import { PAYMENT_METHOD_LIST } from '~/constant';
+import { PAYMENT_METHOD_LIST, LIKE_CO_CLOUD_FN_BASE } from '~/constant';
 
 export default {
   components: {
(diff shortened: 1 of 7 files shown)


src/plugins/geoip.server.js
```diff
@@ -1,14 +1,38 @@
...
   const cacheServerName = req.headers['x-forwarded-server'] || '';
-  if ((ipCountry && ipCountry !== 'ZZ') || ipCity || cacheServerName) {
-    const isHK =
-      ipCountry === 'HK' ||
-      ipCity === 'hong kong' ||
-      cacheServerName.includes('HKG');
-    store.dispatch('setIsHK', isHK);
+  const browserLanguage = detectHKBrowserLang(req.headers['accept-language']); // browser language workaround
+  // Only the cache server of the HKG region causes problem to detect people from HK
+  if (!cacheServerName.includes('HKG')) {
```
williamchong007: Maybe we can just totally ignore the `cacheServerName`  in this whole logic, and just detect if the country is `ZZ` (which means the geoip has invalid value) ?

DarkSorrow: The thing is the ZZ is not an invalid value, it's just saying that the geoip server can't find the country this ip belong to. In the end even if you check it again with the endpoint provided it will give you back a ZZ again.
In this use case we just want to detect HK and the only reason when we aren't sure if it's an HK user because the geoip might be taken from a cache is when the HKG is present in the cacheServer which is why i think it's better to keep it there. The HKG is cache for user in the hong kong area only from what i understood, so with that in mind the only one that needs to get a call to make sure they aren't from HK is those having it, the others are sure not to be HK user so we'll avoid all the other webcall from cached user outside the HK cache area

src/pages/civic/index.vue
```diff
@@ -385,13 +385,28 @@ export default {
...
     };
   },
-  mounted() {
-    if (this.getIsHK) {
+  async beforeMount() {
+    let isHK = this.getIsHK;
+    if (isHK === undefined) {
+      isHK = false; // Default not from HK
+      try {
+        const { data: geoData } = await this.$axios.get('/api/civic/geoip');
```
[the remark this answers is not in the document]
DarkSorrow: Ok no problem, i'll be home in a few hours i'll do that by then. Sorry i wasn't sure where those were before

src/plugins/geoip.server.js
williamchong007: But the problem here is that we could be accessing the website either with or without firebase cdn, and the latter one almost always return `ZZ`. If the `ZZ` returned is due to using CDN, the next step we would do is calling the geoip API above without CDN, which should yield us something other than `ZZ` (hopefully). 

So ignoring cache region, and just check for `ZZ` as the "check again" flag seems more simple in case we want to check for region other than HK later.

DarkSorrow: For checking region other than HK yes. I was just talking about what is being done now but it's fine for me to change it
So tonight i'll check if it's (ipCountry && ipCountry !== 'ZZ') then i check the variables otherwise i will make the call on the frontend

williamchong007: Yes these 2 should be all the change needed, thanks a lot for the work

src/pages/civic/index.vue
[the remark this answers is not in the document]
DarkSorrow: the const LIKE_CO_CLOUD_FN_BASE seems to be the same link you gave me want me to add a new one or i use that one?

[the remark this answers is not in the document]
williamchong007: Oops
if they are identical, just import the existing one will do

DarkSorrow:
I'm just not sure if the end of the API link is correct, when you deploy it takes the API infront?

PR #233, opened 2020-03-04
♻️ Write geoip into separated function and use maxwind db

williamchong007:
This fixes strange cors error by putting geoip under same origin


functions/geoipHttp/index.js
```diff
@@ -0,0 +1,22 @@
...
+
+const app = express();
+app.use(helmet());
+app.get('/api/geoip', (req, res) => {
+  res.setHeader('Cache-Control', 'no-cache, no-store, must-revalidate');
+  try {
+    const result = geoip.lookup(
+      req.headers['fastly-client-ip'] || requestIp.getClientIp(req)
+    );
+    res.json({ country: result.country });
```
nnkken: `result` could be `null`

PR #243, opened 2020-03-23
Update FPS ID & PayMe QR code

nwingt:


nwingt:
This page is mainly for HK people who can read Chinese
...
Actually we are lazy to translate the page

PR #255, opened 2020-05-14
✨ Add default follow IDs

williamchong007:


src/server/api/routes/users.js
```diff
@@ -127,14 +131,17 @@ router.post('/users/login', async (req, res, next) => {
...
-      { merge: true }
-    );
+
+    const payload = {
+      user: userData,
+      accessToken,
+      refreshToken,
+    };
+    if (isNew) {
+      payload.followedUsers = FieldValue.arrayUnion(...DEFAULT_FOLLOW_IDS);
```
nnkken, line 141: `isNew` means `userDoc` does not exist, does array union work in this case, and is it needed?

src/server/api/constant/index.js
```diff
@@ -7,7 +7,16 @@ const AUTH_COOKIE_OPTION = {
   httpOnly: true,
 };
 
+const DEFAULT_FOLLOW_IDS = [
+  'hi176-matters',
+  'standnews',
+  'inmediahknet',
+  'hkcitizennews',
+  'ckxpress',
+];
```
nwingt, line 16: Perhaps sort the array items?

src/server/api/routes/users.js
nwingt: https://cloud.google.com/firestore/docs/manage-data/add-data#data_types

Seems plain array should be fine? 🤨

williamchong007: It works for non exist, but yes actually not needed

PR #257, opened 2020-05-27
Support super like

williamchong007:


src/server/api/routes/reader/index.js
```diff
@@ -16,6 +19,18 @@ const router = Router();
 router.use(follow);
 router.use(bookmark);
 
+function filterSuperLikeList(list) {
+  return list.map(l => {
+    const { id, likee, ts, url } = l;
```
nwingt, line 24: Can also pass `liker` for UI?

PR #263, opened 2020-06-18
✨ Invalidate user session if some scope is missing in current token

williamchong007:


src/server/api/util/api.js
```diff
@@ -132,7 +133,7 @@ const getOAuthURL = ({ state, isRegister, from, referrer }) => {
   const qsPayload = {
     client_id: LIKE_CO_CLIENT_ID,
     redirect_uri: OAUTH_REDIRECT_URI,
-    scope: 'profile email read:like.info read:civic_liker write:civic_liker',
+    scope: OAUTH_SCOPE_REQUEST.join(' '),
```
nnkken, line 136: `OAUTH_SCOPE_REQUEST` or `OAUTH_SCOPE_REQUIRED`? Seems the original one is `OAUTH_SCOPE_REQUIRED`

williamchong007: `REQUEST` because we have a new feature that auto spread scope without `read` or `write` into both `read` and `write`

williamchong007:
> side able to detect the error and r

I am taking the lazy approach of making the client logged out if scope is not enough. A re-login should automatically reauth with enough scope

PR #266, opened 2020-07-10
🐛 Fix some unfollowed users were not listed in api

williamchong007:


src/server/api/routes/reader/index.js
```diff
@@ -36,16 +36,26 @@ async function getFollowedUserListInfo(req) {
...
+    apiUnfollowed.forEach(u => {
+      userSet.delete(u.id);
+      unfollowedUserSet.add(u);
+    });
   }
-  return { followedUsers: Array.from(userSet), unfollowedUsers };
+  return {
+    followedUsers: Array.from(userSet),
+    unfollowedUsers: Array.from(unfollowedUserSet),
+  };
```
nwingt, line 58: Perhaps sorting the results by Liker ID in asc is better

PR #270, opened 2020-07-17
🚸 Add civic register CTA for civic liker in grace period

williamchong007:


src/pages/settings/civic.vue
```diff
@@ -59,7 +59,13 @@
               )
                 | {{ buttonText }}
         br
+        NuxtLink.btn.btn--plain.btn--auto-size.text-12(
```
nwingt, line 62: L57 `class="buttonClass"`

src/pages/settings/civic.vue
```diff
@@ -59,7 +59,14 @@
               )
                 | {{ buttonText }}
         br
+        NuxtLink(
+          class="buttonClass"
```
nwingt, line 63: No, I mean L57 should be `:class="buttonClass"` not `class="buttonClass"`

PR #292, opened 2020-09-07
✨ Add about page

nwingt:


williamchong (commented):
should the paragraph be in i18n or is this a hack?

nwingt:
This page is just for Google Ads Grant applications, don't need translation for now

williamchong (approved):

williamchong (approved):

PR #295, opened 2020-09-11
✨ Add support to superlike short id and new url

williamchong:


src/components/ContentCardWrapper.vue
```diff
@@ -99,9 +103,10 @@ export default {
       return !this.internalTitle || !this.internalLikeCount;
     },
     href() {
-      return this.superLikeId
-        ? getSuperLikeRedirectLink(this.superLikeId)
-        : this.internalUrl;
+      if (this.superLikeShortId)
+        return getSuperLikeRedirectLink(this.superLikeShortId);
```
nwingt, line 107: if the return statement is on the next line, perhaps consider adding brackets

nwingt (changes requested):

PR #296, opened 2020-09-24
💄 Update feed tabs

nwingt:


williamchong:
Currently the feature feed turn into global super like feed, after user logged in and is a super liker. Do we want to show the global super like feed also?

nwingt:
Thanks for reminding me. Updated.

williamchong:
But if you remove `getFetchSuggestArticlesApi ` we will be show global super like feed to guest in frontpage, which might contain spam.
My previous comment was more on about should we show the global super like feed to logged in super likers.

nwingt:
Confirmed with @joshkiu. Super Like global feed is not shown to signed-in users

williamchong (approved):

PR #297, opened 2020-09-24
Fix possible flashing & API hotfix 

nwingt:


nnkken (approved):

williamchong (approved):

williamchong:
which part use latest?

nwingt:
It is the app. I forgot about it. 💩

PR #299, opened 2020-09-29
♻️ Refactor error handler to allow 4xx error to be emit back

williamchong:
Return original 4xx error back to client

### DIFF
--- a/src/server/api/middleware/error.js
+++ b/src/server/api/middleware/error.js
@@ -1,0 +1,16 @@
+function handleRestfulError(req, res, next, err) {
+  if (
+    err.response &&
+    err.response.status >= 400 &&
+    err.response.status <= 499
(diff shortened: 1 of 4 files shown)


src/server/api/middleware/error.js
```diff
@@ -0,0 +1,16 @@
+function handleRestfulError(req, res, next, err) {
+  if (
+    err.response &&
+    err.response.status >= 400 &&
+    err.response.status <= 499
+  ) {
+    res.set('Content-Type', 'text/plain');
+    res.status(err.response.stautus).send(err.response.data);
```
nwingt, line 8: stautus -> status

nwingt (changes requested):

nwingt (approved):

PR #314, opened 2020-10-23
⚡️  Various performance improvement

williamchong:
I am not sure if preinstall script would work in firebase, would be nice if extract css plugin actually make a release soon

### DIFF
--- a/functions/package-lock.json
+++ b/functions/package-lock.json
@@ -3282,15 +3282,6 @@
         "@types/mime": "*"
       }
     },
-    "@vimeo/player": {
-      "version": "2.9.1",
-      "resolved": "https://registry.npmjs.org/@vimeo/player/-/player-2.9.1.tgz",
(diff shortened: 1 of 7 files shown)


src/plugins/ui-plugin.client.js
```diff
@@ -9,17 +8,6 @@ import {
...
-  window &&
-  'IntersectionObserver' in window &&
-  'IntersectionObserverEntry' in window &&
-  'intersectionRatio' in window.IntersectionObserverEntry.prototype;
-
-Vue.use(VueLazyLoad, {
-  lazyComponent: true,
-  observer: hasIntersectionObserverSupport,
-});
-
```
nwingt, line 22: We are still using `vue-lazyload`, since it does support SSR, so the component is globally registered here
https://github.com/likecoin/liker-land/blob/cd062e60f72f035b2467fe2f3a76ed8ce2609ba0/src/components/ContentCardWrapper.vue#L2-L20

nwingt (changes requested):

williamchong:
Force pushed to keep vue-lazy, and since hacking in the latest version of extract css is hell to deal with, I gave up for now and reverted related commits

nwingt (approved):

PR #318, opened 2020-10-29
Add portfolio page

nwingt:


src/components/Button/Button.vue
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  &:active > div {
+    background-color: #0000001a;
+  }
+
+  &__title {
+    padding: 0 8px;
+
+    font-family: Source Sans Pro, Arial, sans-serif;
```
williamchong, line 123: Should `Source Sans Pro, Arial, sans-serif` be in some kind of variable, instead of appearing in different components?

src/components/Button/ButtonGroup.vue
```diff
@@ -0,0 +1,53 @@
+<script>
+export default {
+  name: 'ButtonGroup',
+  render(createElement) {
+    const children = [];
+    this.$slots.default.forEach(node => {
+      if (node.tag) {
```
williamchong, line 7: In what case a node would have no tag?

src/components/SuperLikeContentCard.vue
```diff
@@ -0,0 +1,247 @@
...
+            this.fetchUserInfo(this.author.user)
+              .then(() => this.updateAuthorInfo())
+              .catch(() => ({}))
+          );
+        } else {
+          this.updateAuthorInfo();
+        }
+        await Promise.all(promises);
+      } finally {
+        this.$emit('fetched');
```
williamchong, line 174: Should the component still emit `fetched` if any of the promise throws?

src/components/SuperLikeContentCard.vue
```diff
@@ -0,0 +1,247 @@
...
+        this.internalCoverSrc = image;
+        this.internalLikeCount = like;
+        // If authorId is not given, we can retrieve from article info
+        if (!this.author.user) {
+          this.author = { user };
+        }
+      }
+    },
+    updateAuthorInfo() {
+      this.author = this.getUserInfoById(this.author.user) || this.author;
```
williamchong, line 200: Would having `authorId` as string data, computed prop based on `authorId` that returns `getUserInfoById `, and a watcher on `authorId` to fetch user data if not exists be a easier to understand structure than this union?

src/components/Button/Button.vue
nwingt: I think we are going to do it in a theme provider component in the future, I think this will do for now

src/components/Button/ButtonGroup.vue
nwingt: plain text

src/components/SuperLikeContentCard.vue
nwingt: I don't know, this component is duplicated from ContentCardWrapper, and you wrote this line:
https://github.com/likecoin/liker-land/blame/258487ea565402c8097f88b6a419b7a737c2355d/src/components/ContentCardWrapper.vue#L177

src/components/SuperLikeContentCard.vue
nwingt: This is for parent component to recalculate view size after loading

src/components/SuperLikeContentCard.vue
williamchong: Maybe we(i) can refactor this part later

williamchong (approved):
I think the svg issue can be optimized later if we don't want to spend too much time on profiling this PR

src/components/Identity/Identity.avatar.vue
```diff
@@ -1,47 +1,34 @@
...
....5 3.1c0 .2-.2.2-.3.1l-1.1-.9c-.1 0-.1-.1-.1-.2l.7-3.2-.6-.5-1.7 2.2c-.1.1-.2.1-.3 0l-1-.8c-.1-.1-.1-.2 0-.3-.3.2 4.2-5.5 4.2-5.5zM15.4 38.3c-.8 0-1.5-.7-1.5-1.5 0-.3.1-.6.2-.8 2.4-3.7 5.1-7.1 8.1-10.3.6-.6 1.5-.6 2.1-.1.6.6.6 1.5.1 2.1-2.9 3.1-5.5 6.4-7.7 9.9-.4.4-.8.7-1.3.7M5.8 60.2c-.8 0-1.5-.7-1.5-1.5v-.4c1.2-5 3-9.8 5.2-14.4.4-.7 1.3-1 2-.7.7.4 1 1.2.7 2-2.1 4.4-3.8 9-5 13.8-.1.8-.7 1.2-1.4 1.2m18.8 68c-.4 0-.8-.2-1.1-.4-13.7-14-21.3-32.4-21.3-51.7 0-.8.7-1.5 1.5-1.5s1.5.7 1.5 1.5c0 18.6 7.3 36.2 20.5 49.6.6.6.6 1.5 0 2.1-.3.3-.7.4-1.1.4m51.3 21.5c-15.4 0-30.2-4.8-42.8-13.9-.7-.5-.9-1.4-.4-2.1s1.4-.9 2.1-.4h.1C47 142 61.2 146.6 76 146.6c18.6 0 36.2-7.3 49.6-20.5.6-.6 1.5-.6 2.1 0s.6 1.5 0 2.1c-14 14-32.4 21.5-51.8 21.5m-46-16.8c-.3 0-.7-.1-1-.3-.5-.4-1-.8-1.4-1.2-.6-.6-.6-1.5-.1-2.1s1.4-.6 2.1-.1c.5.4.9.8 1.4 1.2.6.5.7 1.5.2 2.1-.3.2-.7.4-1.2.4m106.7-16.2c-.8 0-1.5-.7-1.5-1.5 0-.3.1-.6.2-.8 7.5-11.5 11.4-24.7 11.4-38.3 0-10-2.1-19.9-6.2-29-.3-.8 0-1.6.8-2 .8-.3 1.6 0 2 .8 4.3 9.5 6.5 19.8 6.5 30.3 0 14.2-4.1 28-11.9 40-.4.2-.8.5-1.3.5m-.7-79.5c-.5 0-1-.2-1.2-.7-1.9-2.9-4.1-5.6-6.4-8.2-.5-.6-.5-1.6.2-2.1.6-.5 1.5-.5 2.1.1 2.4 2.7 4.7 5.5 6.7 8.5.5.7.3 1.6-.4 2.1-.4.2-.7.3-1 .3"
```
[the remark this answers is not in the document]
nwingt: Reference for the halo in Medium
image

I prefer using inline SVG for controlling fill by `currentColor`, since  can't do that, I think I am going to split this part to another component, see if that is acceptable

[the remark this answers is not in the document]
williamchong: I think doing sth like https://github.com/likecoin/liker-land/pull/314/commits/737cc71ce7c04bbf546f098ea19e277f1c9f87ae will make lighthouse happy, not sure if it does make it faster for ux though

PR #322, opened 2020-11-02
⬆️ Upgrade extract-css-chunks-webpack-plugin for css preload

williamchong:
This would conflict wit #318 so maybe we should handle either one first


nwingt (approved):

nwingt:
@williamchong Maybe you merge this first then I rebase #318 onto this?

PR #323, opened 2020-11-03
Fix stack layout & lazy load

nwingt:


src/components/SuperLikeContentCard.vue
```diff
@@ -3,15 +3,20 @@
     class="super-like-content-card"
     @show="fetchContent"
   )
+    Card(v-if="isLoading")
+      Placeholder(style="width: 60%;height: 16px")
+      Placeholder(style="width: 100%;height: 16px;margin-top: 12px")
+      Placeholder(style="width: 40%;height: 16px;margin-top: 8px")
```
williamchong, line 9: why 3 placeholder instead of one?

nwingt: Each placeholder is a rectangle
Screenshot 2020-11-03 at 6 04 13 PM

williamchong (approved):

PR #331, opened 2020-11-12
Portfolio page improvements

nwingt:


src/locales/zh-Hant.json
```diff
@@ -270,7 +270,10 @@
       "Works": "作品",
       "SuperLikes": "已 Super Like 內容"
     },
-    "EmptyLabel": "沒有作品"
+    "EmptyLabel": "沒有作品",
+    "Og": {
+      "Title": "{name}的個人專頁"
```
williamchong, line 275: Having a space between `{name}` and `的個人專頁` should be more robust to different names

src/pages/_id/index.vue
```diff
@@ -189,6 +189,7 @@ export default {
   },
   mounted() {
     this.refreshBookmarkList();
+    this.fetchReaderIndex();
```
williamchong: These 2 promises should be in a `Promise.all` ? (Not sure about if these 2 can merge with the one below)

nwingt: Hmm, but these 2 calls do not have order dependency

williamchong (approved):

PR #346, opened 2020-12-04
🩹 Filter out self SL in portfolio SL tab

nwingt:


williamchong (approved):
Not sure what is the context for filtering

nwingt:
to hide the creator' contents in the Super Liked tab

PR #355, opened 2020-12-28
📈 Add civic liker funnel events

williamchong:


nwingt (commented):
I don't think it is a good practice to emit event outside a page component, this component is probably going to be reused in the future, perhaps using `@emit` with callbacks of the corresponding page

nwingt (approved):

PR #356, opened 2021-01-04
🐛 Fix cannot resume canceled subscription

williamchong:


nwingt (approved):
Using DELETE for resuming is a bit weird though

williamchong:
couldn't find a good existing endpoint just for POSTing a resume, so i guess DELETE the cancel still make some sense...

nwingt:
Hotfix is ok

nwingt (approved):

williamchong:
@nwingt on second thought maybe the change will affect those actually expired, so I am splitting the `isCivicLikerRenewalPeriod` and `willCancel` case

PR #361, opened 2021-01-06
🐛 Fix CL welcome dialog not shown in the first time

nwingt:


williamchong (approved):
Didn't know it would make a difference...

PR #362, opened 2021-01-07
🌐 Sync like.co language preferences

nwingt:
Depends on https://github.com/likecoin/likecoin-api-public/pull/213


williamchong (approved):
`apiFetchUserPreferences` is not used any where right?

nwingt:
Yes

PR #365, opened 2021-01-08
Fix CL2.0 sub flow

nwingt:
Depends on https://github.com/likecoin/like-co/pull/1489

### DIFF
--- a/src/components/CivicSubscriptionView/CivicSubscriptionView.vue
+++ b/src/components/CivicSubscriptionView/CivicSubscriptionView.vue
@@ -550,6 +550,8 @@ export default {
       try {
         await Promise.all(promises);
       } catch (err) {
+        // eslint-disable-next-line no-console
+        console.error(err);
         this.$router.replace({ name: 'settings-civic' });
       }
       // Set default to  1 for new subscription
--- a/src/components/CivicSubscriptionView/CivicSubscriptionView.vue
+++ b/src/components/CivicSubscriptionView/CivicSubscriptionView.vue
@@ -187,6 +187,7 @@
         </div>
 
         <div
+          v-if="!isUserCurrentCivicV1"
           class="mt-24 text-16 font-500"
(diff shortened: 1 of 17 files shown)


williamchong (approved):

src/components/CivicSubscriptionView/CivicSubscriptionView.vue
```diff
@@ -469,8 +511,6 @@ export default {
       try {
         await Promise.all(promises);
       } catch (err) {
-        // eslint-disable-next-line no-console
-        console.error(err);
```
williamchong, line 473: Why remove, do we expect any error now?

src/util/api/index.js
```diff
@@ -61,9 +62,20 @@ export const getUpdateReaderBookmarkAPI = url =>
...
 export const getFetchFollowedSuperLikeApi = () =>
   `/api/reader/superlike/followed`;
-export const getOAuthRegisterAPI = (from = '', referrer = '') =>
-  `/api/users/register?from=${from}&referrer=${encodeURIComponent(referrer)}`;
-export const getOAuthLoginAPI = () => '/api/users/login';
+export const getOAuthRegisterAPI = (
+  language = 'zh',
+  from = '',
+  referrer = ''
+) =>
```
williamchong, line 69: Would using named param better? These 3 params are all optional, and doesn't have a logical relationship that one must be placed earlier than the others.

williamchong (changes requested):

src/components/CivicSubscriptionView/CivicSubscriptionView.vue
nwingt: Updated

src/util/api/index.js
nwingt: Updated

williamchong (approved):

PR #382, opened 2021-01-26
🐛 Fix experiment plugin code for gtag

williamchong:
Ref: https://stackoverflow.com/questions/50686295/gtag-js-optmize-server-side-implementation-not-working

Not sure if we should be using variant name or indexes though


nwingt:
ref to the answer, is it required to input all variants?

williamchong:
1 variant for each experiment I suppose

nwingt (approved):

PR #386, opened 2021-02-04
♻️ Use localstorage for post auth route storing

williamchong:


nwingt:
Need to test?

williamchong:
> Need to test?

Tested logging in from Civic Liker page, seems no problem

williamchong:
Maybe need to test if iOS/OSX safari's ITP complicate things here, I just tested desktop chrome

nnkken (approved):

nwingt (approved):

PR #387, opened 2021-02-08
Minor improvement for reg flow

nwingt:


williamchong (approved):

nwingt:
@williamchong Change `isDisabledNav` default to false

williamchong (approved):

PR #395, opened 2021-03-05
Show default creator pitch if not set 

nwingt:


williamchong:
Does this affect og description too?

nwingt:
@williamchong Updated

williamchong (approved):

PR #402, opened 2021-03-11
🐛 Allow civic liker in past_due or grace period to enter civic setting page

williamchong:
- Add a page that directly links to stripe billing page so that we can set this url in url/customer support message
- Allow expired user to enter civic setting page (which default to show v1) so that that won't be in an infinite loop in civic pages
- Show a message if we detect the user is `past_due` instead of `cancel` or no subscription, so that billing method can be updated

Known issue:
- Would show v1 option page even for v2 user
- After updating billing method, invoice payment retry is not auto triggered. We still won't treat v2 user as v2 user, and they cannot enter v2 panel until a payment retry is triggered and success
civic

### DIFF
--- a/src/components/CivicSettingsV1.vue
+++ b/src/components/CivicSettingsV1.vue
@@ -37,6 +37,10 @@
                 )
                   | {{ $t('SettingsCivicPage.editPaymentMethod') }}
               .settings-civic-page__billing-summary-row.liker-comparison-card__b--mx
+                v-if=isPastDueUser
+                .settings-civic-page__billing-summary-row-value.text-danger
+                  | {{ $t('SettingsCivicPage.paymentExpired') }}
+              .settings-civic-page__billing-summary-row.liker-comparison-card__b--mx
                 label.settings-civic-page__billing-summary-row-label
                   | {{ $t(`SettingsCivicPage.billingSummary.${getUserSubscriptionInfo.willCancel ? 'cancel' : 'nextBilling'}Date`) }}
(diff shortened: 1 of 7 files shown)


src/components/CivicSettingsV1.vue
```diff
@@ -215,10 +220,14 @@ export default {
...
-      if (this.getUserIsCivicLikerPaid) {
+    async fetchSubscriptionInfo({ force = false } = {}) {
+      if (this.getUserIsCivicLikerPaid || this.getUserShouldRenewCivic) {
         try {
-          const { willCancel } = await this.fetchUserSubscriptionInfo();
+          if (force || !this.getUserSubscriptionInfo) {
+            await this.fetchUserSubscriptionInfo();
+          }
+          const { willCancel, status } = this.getUserSubscriptionInfo;
+          this.isPastDueUser = status === 'past_due';
```
nwingt, line 230: turn into getter?

nwingt (approved):

src/components/CivicSettingsV1.vue
williamchong: Easier to unify in case later we want to do `fetchUserSubscriptionInfo()` action outside panel, maybe in the parent page

williamchong: You mean turn `isPastDueUser` into getter?

williamchong: Will open another PR

PR #406, opened 2021-03-24
Unify portfolio CTA route & fix avatar fallback src

nwingt:


src/pages/_id/index.vue
```diff
@@ -255,22 +255,10 @@ export default {
       return this.supportingAmount > 0;
     },
     ctaToRoute() {
-      if (this.getUserIsCivicLikerV2) {
-        if (this.isSupportingCreator) {
-          return undefined;
```
williamchong, line 260: So there is no undefined case now?

williamchong (approved):

src/pages/_id/index.vue
nwingt: Updated

williamchong: How about the `getUserIsCivicLikerV2` checking?

PR #415, opened 2021-03-27
Update civic page creators section UI & API

nwingt:
Make creators in the following section clickable
image

Let merge to feature/civic first if need tweaking on API for suggested contents


src/pages/civic/_from.vue
```diff
@@ -31,6 +34,31 @@ export default {
...
+      'getArticleInfoByReferrer',
+      'getUserInfoById',
+    ]),
+    creators() {
+      const creators = [];
+      if (this.creator) {
+        creators.push(this.creator);
+      }
+      creators.push(
+        ...this.contents.map(
```
williamchong, line 55: Would `.concat()` be simpler?

williamchong (changes requested):

src/pages/civic/_from.vue
nwingt: Changed to `.unshift` then

williamchong (approved):

PR #419, opened 2021-04-08
💄 Update open graph for civic/_id page

WeiJun0827:


src/pages/civic/_from.vue
```diff
@@ -68,5 +68,42 @@ export default {
...
+          property: 'og:description',
+          content: description,
+        },
+        {
+          hid: 'og:image',
+          property: 'og:image',
+          content: 'https://liker.land/images/og/civic-v2.png',
+        },
+      ],
+      link: [{ rel: 'canonical', href: `${this.$route.path}` }],
```
williamchong, line 105: canonical might not be path since we have querystring `from`, please remove this line for now since it is irrelevant to this PR

williamchong (changes requested):

PR #422, opened 2021-04-13
Various fixes on /civic and others

nwingt:
# Split settings header
image
## Fix
image

# Cropped avatar of suggested creators
image

## Fix
Allow controlling avatar outline extrusion
image

image

# Show placeholder when loading suggested creators
image

# FIx undefined suggested creators
image

### DIFF
--- a/src/components/CivicLikerPage/sections/SuggestedCreatorsSection.vue
+++ b/src/components/CivicLikerPage/sections/SuggestedCreatorsSection.vue
@@ -10,6 +10,7 @@
         preset="creator"
         :contents="contents"
         :is-waterfall-layout="false"
+        :is-loading="contents.length === 0"
       )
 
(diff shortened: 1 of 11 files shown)


src/components/CivicLikerPage/sections/SuggestedCreatorsSection.vue
```diff
@@ -10,6 +10,7 @@
         preset="creator"
         :contents="contents"
         :is-waterfall-layout="false"
+        :is-loading="contents.length === 0"
```
williamchong, line 13: Does there exists a case `contents` is really of length 0?

nwingt: I think not, contents should be covered by suggested API if personal API return empty

williamchong (approved):

PR #424, opened 2021-04-14
💄 Support dynamically generated og image feature

WeiJun0827:


williamchong (approved):

src/pages/_id/civic/index.vue
```diff
@@ -147,7 +147,7 @@ export default {
         {
           hid: 'og:image',
           property: 'og:image',
-          content: 'https://liker.land/images/og/civic-v2.png',
+          content: `https://static.like.co/liker-og-image/${this.creator.user}`,
```
nwingt, line 150: Maybe refactor this line into a function?

nwingt (changes requested):

src/pages/_id/civic/index.vue
WeiJun0827: Updated, please check.

src/util/api/index.js
```diff
@@ -131,3 +131,5 @@ export const getArticleDetailAPI = url =>
 
 export const updateProfile = () => `/api/users/self/update`;
 export const userPreferences = () => `/api/users/preferences`;
+
+export const getOgImage = id => `https://static.like.co/liker-og-image/${id}`;
```
nwingt, line 135: Suggest to rename this func for more specific use case
```suggestion
export const getLikerOgImage = id => `https://static.like.co/liker-og-image/${id}`;
```

nwingt (changes requested):

src/util/api/index.js
WeiJun0827: OK! Done~

nwingt (approved):

nwingt (approved):

PR #433, opened 2021-05-03
🚑 Fix Super Like grid item not update

nwingt:
Fix item not re-render after items update


williamchong:
Should both key be `item.superLikeID` ?

nwingt:
@williamchong Not really, only update inner div should be enough

williamchong (approved):

PR #443, opened 2021-06-25
✨ Change getApp page background color to grey

shellyYG:


src/pages/getapp.vue
```diff
@@ -59,6 +59,6 @@ export default {
 
 <style lang="scss">
 .getapp-page {
-  background: white;
+  background: #f7f7f7;
```
nwingt, line 62: I think you can remove the CSS here, as I rmb #f7f7f7 is the default color

nwingt (changes requested):

src/pages/getapp.vue
```diff
@@ -59,6 +59,6 @@ export default {
...
 }
 </style>
```
nwingt, line 64:
```suggestion
```

nwingt (changes requested):

src/pages/getapp.vue
shellyYG: true, revert now

nwingt (approved):

PR #480, opened 2022-03-16
🔥 Remove About page and related  buttons

AuroraHuang22:
截圖 2022-03-16 上午11 18 47

### DIFF
--- a/src/pages/about.vue
+++ b/src/pages/about.vue
@@ -1,17 +1,7 @@
-<template lang="pug">
-  .about-page
-    PageHeader
-      template
-        SiteNavBar.text-white.bg-like-green
-
-    main.page-content
-      AboutPage
-</template>
-
 <script>
 export default {
-  mounted() {
-    window.location.href = 'https://about.like.co/';
+  fetch(context) {
+    context.redirect('https://about.like.co/');
   },
 };
 </script>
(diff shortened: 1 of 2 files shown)


src/pages/about.vue
```diff
@@ -9,18 +9,9 @@
...
-
-import { CrispMixinFactory } from '~/mixins/crisp';
-
 export default {
-  components: {
-    AboutPage,
-    PageHeader,
-    SiteNavBar,
+  created() {
+    this.$router.push({ name: 'index' });
```
nwingt, line 14: Should be redirected to about.like.co??

src/pages/about.vue
```diff
@@ -9,18 +9,9 @@
...
-
-import { CrispMixinFactory } from '~/mixins/crisp';
-
 export default {
-  components: {
-    AboutPage,
-    PageHeader,
-    SiteNavBar,
+  mounted() {
+    window.location.href = 'https://about.like.co/';
```
nwingt, line 14: Please use `fetch(context)` instead of `mounted()` life cycle with `context.redirect(URL)` and remove everything else

nwingt (changes requested):

src/pages/about.vue
AuroraHuang22: wow, so cool 👀, It's much simpler ~

nwingt (approved):

PR #488, opened 2022-06-23
📦 Migrate Tailwind v0 to v3

AuroraHuang22:
- Update tailwind to v3 https://tailwindcss.com/blog/tailwindcss-v3-1


src/components/SlidingMenu.vue
```diff
@@ -176,7 +176,8 @@ html[sliding-menu='opened'] {
   width: $sliding-menu-width;
 
   @apply fixed;
-  @apply pin-t;
+  /* @apply inset-x-0; */
```
nwingt, line 179:
```suggestion
```

src/components/TabBarItem.vue
```diff
@@ -90,7 +90,7 @@ export default {
     }
 
     &__content {
-      @apply bg-current;
+      @apply bg-inherit-color;
```
nwingt, line 93: Should be the same `bg-current` which is `background: currentColor`

nwingt (approved):

nwingt:
@williamchong Any comments?

src/package.json
```diff
@@ -45,30 +47,45 @@
...
-    "vue-gtag": "^1.10.0",
+    "vue-gtag": "^1.16.1",
     "vue-i18n": "^8.9.0",
     "vue-lazyload": "^1.2.6",
+    "vue-property-decorator": "^9.1.2",
+    "vue-server-renderer": "^2.6.14",
     "vue-vimeo-player": "0.0.9",
-    "vuex": "^3.1.2"
+    "vuex": "^3.1.2",
+    "vuex-class": "^0.3.2"
```
williamchong, line 72: `vue` and `vue-server-renderer` (and vue-template-compiler)should have exact same version
in fact I usually just don't declare them in package.json and let them obey nuxt defined versions

williamchong (changes requested):

functions/tailwind.config.js
```diff
@@ -0,0 +1,877 @@
+/* eslint-disable prettier/prettier */
```
nwingt, line 1: Is it necessary to add tailwind config in functions?

AuroraHuang22: functions also has a nuxt.config.js file, this file requires tailwind.config as the theme.

nwingt: Firebase will run the following script to copy the file from src
https://github.com/likecoin/liker-land/blob/42aabdb28903175b2700d7c1ef7170a3d263b262/firebase.json#L5

williamchong (approved):

PR #496, opened 2022-07-18
Add about NFT page

nwingt:


src/components/AboutNFTPage/NFTWidgetSection.vue
```diff
@@ -0,0 +1,277 @@
...
+        </div>
+      </div>
+      <div
+        class="
+          nft-widget-graph__description
+          mt-[16px]
+          laptop:mt-[0px] laptop:max-w-[320px] laptop:mb-[70px] laptop:ml-[32px]
+          leading-[22px]
+        "
+        v-html="$t('about_nft_page_nft_widget_description')"
```
williamchong, line 162: Prefer i18n path than `v-html` if possible

williamchong (approved):

src/components/AboutNFTPage/NFTWidgetSection.vue
nwingt: There are some bold texts in between the translated content, it seems `v-html` is the only [option](https://kazupon.github.io/vue-i18n/guide/formatting.html#html-formatting) without using the complicated interpolation

src/locales/en.json
```diff
@@ -1,4 +1,43 @@
...
+  "about_nft_page_join_the_community_title": "加入社群",
+  "faq_list": [
+    {
+      "q": "Why publish articles as Writing NFT?",
+      "a": "An individual’s content portfolio can be viewed as part of a reliable decentralized identity in Web3. It’s a record of individual authorship. Writing NFT can guarantee the immutability of content and later verifiable with proof of ownership and property rights."
+    },
+    {
+
+      "q": "How to get started with Writing NFT?",
+      "a": "There are three ways, 1) if you are a WordPress user, you can directly install the free LikeCoin plugin from the WordPress plugins. 2) via NFT Portal, you can mint NFT from any URLs, and it will generate an NFT listing page for you. 3) via the partnered website. There are several media/ blogging platforms that have integrated with LikeCoin. {{list}}"
```
AuroraHuang22, line 24:
```suggestion
      "a": "There are three ways, 1) if you are a WordPress user, you can directly install the free LikeCoin plugin from the WordPress plugins. 2) via NFT Portal, you can mint NFT from any URLs, and it will generate an NFT listing page for you. 3) via the partnered website. There are several media/ blogging platforms that have integrated with LikeCoin."
```
Not sure why we need `{{list}}`. Will these platforms be listed in `{{list}}`?

AuroraHuang22 (approved):

PR #505, opened 2022-07-28
💄 QA following adjustments

AuroraHuang22:


nwingt (approved):
Suggest to refactor into a Tooltip component

src/mixins/nft.js
```diff
@@ -126,8 +128,20 @@ export default {
...
+          data.avatar ||
+            `https://avatars.dicebear.com/api/identicon/${addr}.svg`
+        );
+        Vue.set(this.civicLikerList, addr, true);
       } catch (error) {
         Vue.set(this.displayNameList, addr, addr);
+        Vue.set(
+          this.avatarList,
+          addr,
+          `https://avatars.dicebear.com/api/identicon/${addr}.svg`
```
nwingt, line 143: Can be refactored as a function

src/mixins/nft.js
```diff
@@ -9,6 +9,8 @@ export default {
       iscnOwnerInfo: {},
       NFTHistory: [],
       displayNameList: [],
+      avatarList: [],
+      civicLikerList: [],
```
nwingt, line 13: suggest using Vuex state for convenient global access

PR #510, opened 2022-07-29
➕ Integrate `likecoin-wallet-connector`

WeiJun0827:


src/constant/index.js
```diff
@@ -53,10 +53,18 @@ export const STRIPE_SDK_URL = 'https://js.stripe.com/v3/';
...
+
+export const LIKECOIN_CHAIN_DENOM = IS_TESTNET ? 'EKIL' : 'LIKE';
+
 export const CIVIC_LIKER_V3_STAKING_ENDPOINT = IS_TESTNET
   ? 'https://likecoin-public-testnet-5.netlify.app/validators'
   : 'https://dao.like.co/validators';
 
+export const LIKECOIN_CHAIN_NFT_LCD = IS_TESTNET
+  ? 'https://node.testnet.like.co/'
+  : 'https:/mainnet-node.like.co/';
```
AuroraHuang22, line 66:
```suggestion
  : 'https://mainnet-node.like.co/';
```

src/package.json
```diff
@@ -47,6 +47,7 @@
     "firestore-store": "^1.0.0",
     "gsap": "^3.10.4",
     "jsonwebtoken": "^8.5.1",
+    "likecoin-wallet-connector": "git://github.com/nwingt/likecoin-wallet-connector.git#41d65d6326430f7c652e2913170070d654f4fc8f",
```
AuroraHuang22, line 50: please also add this package to [functions](https://github.com/likecoin/liker-land/blob/master/functions/package.json)

src/constant/index.js
```diff
@@ -83,3 +91,20 @@ export const LIKECOIN_NFT_CAMPAIGN_ITEMS = IS_TESTNET
...
+  coinDenom: LIKECOIN_CHAIN_DENOM,
+  coinMinimalDenom: LIKECOIN_CHAIN_MIN_DENOM,
+  coinDecimals: 9,
+  bech32PrefixAccAddr: 'like',
+  bech32PrefixAccPub: 'likepub',
+  bech32PrefixValAddr: 'likevaloper',
+  bech32PrefixValPub: 'likevaloperpub',
+  bech32PrefixConsAddr: 'likevalcons',
+  bech32PrefixConsPub: 'likevalconspub',
+};
```
williamchong, line 110: Just use a separated `network.js` so that we can just reuse keplr configs?

src/store/modules/getters/wallet.js
```diff
@@ -0,0 +1,3 @@
+export const getAddress = state => state.address;
+export const getSigner = state => state.signer;
+export const getConnector = state => state.connector;
```
williamchong, line 3: When would we need to get the connector?

WeiJun0827: We only use the connector in `connectWallet()` to call it's `openConnectWalletModal()`, not sure if we want to store it in the wallet store. Was thinking of declaring `connector` locally in the `/actions/wallet.js`, but we need `commit` param of the `connectWallet()` , so cannot declare `connector` outside of `connectWallet()`.
Should I remove the getter and mutation-type relative with `connector`?

src/constant/index.js
WeiJun0827: sure

src/store/modules/getters/wallet.js
williamchong: hmm maybe keep it for now and see if we use it later

WeiJun0827: OK, then there is no pending issues for now.

PR #512, opened 2022-07-30
💬 Rename Owner to Collector & Update campaign items

AuroraHuang22:


src/pages/nft/class/_classId.vue
```diff
@@ -154,7 +154,7 @@
                 'mb-[16px]',
                 'desktop:m-0',
               ]"
-              title="Owners"
+              title="Collectors"
```
nwingt, line 157: Perhaps add to translation file

AuroraHuang22: I just started to refactor this page, I will move the rest text to locales asap.

nwingt (approved):

PR #518, opened 2022-08-01
🐛 Sort nft owner by desc count

williamchong:


nwingt (commented):
Can move to nft mixin

nwingt (approved):

PR #520, opened 2022-08-01
♻️ [Details Page] Refactor sections

AuroraHuang22:


src/components/NFTPage/EventList.vue
```diff
@@ -0,0 +1,93 @@
...
+              </Label>
+            </td>
+            <td>
+              <Label>{{ event.price }}</Label>
+            </td>
+            <td><Label text="-" /></td>
+            <td>
+              <LinkV2 :to="`/${event.toWallet}`">
+                <Label class="break-all">{{
+                  displayNameList[event.toWallet] | ellipsis
```
nwingt, line 40: Suggest to create a computed property e.g. `populatedEvents` that loop through event array by `array.map` and add a `displayName` key to event array item then return the populated event list, so that no need to pass displayNameList as props

src/components/NFTPage/CollectorList.vue
```diff
@@ -0,0 +1,73 @@
...
+    <template #titleIcon>
+      <slot name="titleIcon" />
+    </template>
+    <template #content>
+      <div class="flex flex-col my-[12px]">
+        <div v-if="ownerCount">
+          <div v-for="o in ownerList" :key="o">
+            <div class="flex items-center justify-between">
+              <LinkV2 :to="`/${o}`">{{ displayNameList[o] || o | ellipsis }}</LinkV2>
+              <Label preset="p6">{{ collectorList[o].length }}</Label>
```
nwingt, line 18: Suggest to create a computed property e.g. `populatedCollectors` that loop through ownerList array by `array.map` and add a `displayName` key and `collectedCount` key to ownerList array item then return the populated ownerList, so that no need to pass displayNameList & collectorList as props

nwingt (changes requested):

src/mixins/nft.js
```diff
@@ -85,6 +85,19 @@ export default {
...
     },
+    populatedEvents() {
+      return this.NFTHistory.map(event => ({
+        ...event,
+        displayName: this.displayNameList[event.toWallet] || event.toWallet,
+      }));
+    },
+    populatedCollectors() {
+      return this.sortedOwnerListId.map(id => ({
+        ...id,
```
nwingt, line 96: no need to spread? 
```suggestion
        id,
```

src/components/NFTPage/CollectorList.vue
```diff
@@ -7,30 +7,28 @@
...
-              :class="[
-                'bg-shade-gray',
-                'h-[1px]',
-                'w-full',
-                'my-[12px]',
-              ]"
-            />
+    <template #content class="flex flex-col my-[12px]">
+      <div v-if="ownerCount">
+        <div v-for="o in ownerList" :key="o.displayName">
```
nwingt, line 14: Should use id as key

nwingt (changes requested):

src/mixins/nft.js
```diff
@@ -85,6 +85,19 @@ export default {
         this.iscnId
       )}%2F1`;
     },
+    populatedEvents() {
+      return this.NFTHistory.map(event => ({
+        ...event,
+        displayName: this.displayNameList[event.toWallet] || event.toWallet,
```
nwingt, line 91: `toDisplayName` is more suitable for this case
```suggestion
        toDisplayName: this.displayNameList[event.toWallet] || event.toWallet,
```

src/mixins/nft.js
```diff
@@ -85,6 +85,19 @@ export default {
...
+    populatedEvents() {
+      return this.NFTHistory.map(event => ({
+        ...event,
+        toDisplayName: this.displayNameList[event.toWallet] || event.toWallet,
+      }));
+    },
+    populatedCollectors() {
+      return this.sortedOwnerListId.map(id => ({
+        id,
+        toDisplayName: this.displayNameList[id] || id,
```
nwingt, line 97: This can keep as `displayName`, since there is no from/to context
```suggestion
        displayName: this.displayNameList[id] || id,
```

nwingt (changes requested):

src/mixins/nft.js
AuroraHuang22: oh, I see what you mean now 🤣

nwingt (approved):

PR #527, opened 2022-08-09
💩 Mitigate ISCN ID double decoding in path on firebase functions

williamchong:
e.g. https://rinkeby.liker.land/nft/iscn/iscn%3A%2F%2Flikecoin-chain%2FDmrE9fnxW-07v7rYLzuJwN73qfNRlhyrJIL5MdAPOIY

### DIFF
--- a/src/layouts/error.vue
+++ b/src/layouts/error.vue
@@ -178,6 +178,17 @@ export default {
     },
   },
   async mounted() {
+    // HACK: to handle firebase function double decoding iscn id, causing route to fail
+    if (this.$route.path.startsWith('/nft/iscn/iscn:/likecoin-chain')) {
+      this.$router.replace({
+        name: 'nft-iscn-iscnId',
+        params: {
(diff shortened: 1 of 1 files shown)


williamchong:
> Probably use router middleware?

You mean extending vue-router?
It doesn't work somehow, I guess it is because the decoded URL contains `iscn:` in path
In vue-router `:` in path means params, so having `/iscn:/....` further breaks the routing...

nwingt (approved):

src/layouts/error.vue
```diff
@@ -178,6 +178,16 @@ export default {
...
   async mounted() {
+    // HACK: to handle firebase function double decoding iscn id, causing route to fail
+    if (this.$route.path.startsWith('/nft/iscn/iscn:/likecoin-chain')) {
+      this.$router.replace({
+        name: 'nft-iscn-iscnId',
+        params: {
+          iscnId: this.$route.path.replace('/nft/iscn/iscn:/', 'iscn://'),
+        },
+      });
+    }
```
nwingt, line 189: Should add return?

PR #538, opened 2022-08-15
🚨 Unify NFT background color type

nwingt:
Need to investigate why `background_color` is not a string

[Example](https://api.like.co/likernft/metadata?class_id=likenft1f9dlnexdt6l4d8u2hh9ccm5g52q46sw2hllk4ntdhf7n6resg00s5hseuj)


williamchong:
Should be fixed by https://github.com/likecoin/likecoin-api-public/pull/332

PR #541, opened 2022-08-16
📦 Replace nuxt-svg-loader with @nuxtjs/svg

nwingt:


williamchong (commented):
if inlining svg means inling in page js bunld, it would make js huge and thus eval time would increase, can we do some profiling?

nwingt:
> if inlining svg means inling in page js bunld, it would make js huge and thus eval time would increase, can we do some profiling?

I tried this approach but it doesn't work
https://github.com/nuxt-community/svg-module#dynamic-imports

williamchong (approved):

PR #546, opened 2022-08-16
🐛 [Details] Display only Writing NFT

AuroraHuang22:
Need to filter out NFTs that are not Writing-NFTs 

截圖 2022-08-16 下午9 51 53

### DIFF
--- a/src/components/NFTPortfolio/Item.vue
+++ b/src/components/NFTPortfolio/Item.vue
@@ -1,5 +1,6 @@
 <template>
   <NuxtLink
+    v-if="isWritingNFT"
     class="mx-auto mb-[5px] break-inside-avoid"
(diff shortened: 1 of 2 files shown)


src/components/NFTPortfolio/Item.vue
```diff
@@ -1,5 +1,6 @@
 <template>
   <NuxtLink
+    v-if="getNFTClassMetadataById(classId)"
```
nwingt, line 3: just use `NFTClassMetadata`?
```suggestion
    v-if="NFTClassMetadata"
```

AuroraHuang22: Maybe `NFTClassMetadata.name` ? since `NFTClassMetadata` is object.

PR #552, opened 2022-08-17
Feat/block api preferences

Tsai-Hsueh-Kuan:
✨ Block no post to preferences

### DIFF
--- a/src/store/modules/actions/user.js
+++ b/src/store/modules/actions/user.js
@@ -1,6 +1,5 @@
 import * as types from '@/store/mutation-types';
 import * as api from '@/util/api';
-
 import {
   updateSentryUser,
   updateCrispUser,
--- a/src/store/modules/actions/user.js
+++ b/src/store/modules/actions/user.js
@@ -1,5 +1,6 @@
 import * as types from '@/store/mutation-types';
 import * as api from '@/util/api';
+
 import {
   updateSentryUser,
   updateCrispUser,
@@ -77,7 +78,7 @@ export async function resumeCanceledSubscription({ dispatch }) {
 
 export async function updatePreferences(
   { dispatch, commit },
-  { locale, creatorPitch } = {}
+  { locale, creatorPitch, loginStatus } = {}
 ) {
   const preferences = {};
   if (locale) {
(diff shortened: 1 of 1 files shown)


src/store/modules/actions/user.js
```diff
@@ -89,6 +89,7 @@ export async function updatePreferences(
     commit(types.USER_UPDATE_USER_INFO, { creatorPitch });
   }
   if (Object.keys(preferences).length) {
-    await this.$api.$post(api.userPreferences(), preferences);
+    // Wait until the correct token can be obtained and then carry loginStatus = true
+    if (loginStatus) await this.$api.$post(api.userPreferences(), preferences);
```
williamchong, line 93: Instead of a new flag, please use the login status in user store
https://github.com/likecoin/liker-land/blob/master/src/store/modules/user.js#L14
to determine if you want to call `updatePreferences` at all or guard this function

williamchong (changes requested):

src/store/modules/actions/user.js
```diff
@@ -77,7 +79,7 @@ export async function resumeCanceledSubscription({ dispatch }) {
 
 export async function updatePreferences(
   { dispatch, commit },
```
nwingt, line 81: Can add getters here, then you can use `getters.getUserInfo` later without import cycle
```suggestion
  { dispatch, commit, getters },
```

nwingt (changes requested):

src/store/modules/actions/user.js
```diff
@@ -91,7 +89,7 @@ export async function updatePreferences(
     commit(types.USER_UPDATE_USER_INFO, { creatorPitch });
   }
   if (Object.keys(preferences).length) {
-    if (userStore.getUserInfo)
+    if (getters.getUserInfo)
```
nwingt, line 92: I just checked `getUserInfo` is an [object](https://github.com/likecoin/liker-land/blob/61e212716a7333ef9c465da3786804cf45aa246f/src/store/modules/user.js#L14), so it is better to use `getUserId` for the condition

nwingt (changes requested):

williamchong (approved):

nwingt (approved):

PR #561, opened 2022-08-23
✨ [Portfolio] Display user stat

WeiJun0827:


src/components/NFTPortfolio/Stat.vue
```diff
@@ -47,6 +47,23 @@ export default {
...
+  },
+  watch: {
+    collectedClassIds() {
+      this.lazyGetAllCollectedPurchaseInfo();
+    },
+  },
+  methods: {
+    ...mapActions(['lazyGetNFTPurchaseInfo']),
+    lazyGetAllCollectedPurchaseInfo() {
+      this.collectedClassIds.forEach(classId =>
```
williamchong, line 63: Set has forEach, do we really need to convert it to array in `collectedClassIds`?

src/store/modules/actions/staticData.js
```diff
@@ -36,6 +36,14 @@ export async function fetchNFTPurchaseInfo({ commit }, classId) {
   return info;
 }
 
+export async function lazyGetNFTPurchaseInfo({ commit, getters }, classId) {
+  let info = getters.getNFTClassPurchaseInfoById(classId);
+  if (!info) {
+    info = await fetchNFTPurchaseInfo.call(this, { commit }, classId);
```
williamchong, line 42: I think for calling other actions from an action, we should use `dispatch `

nwingt, line 42:
```suggestion
export async function lazyGetNFTPurchaseInfo({ commit, getters, dispatch }, classId) {
  let info = getters.getNFTClassPurchaseInfoById(classId);
  if (!info) {
    info = await dispatch('fetchNFTPurchaseInfo', classId);
```

nwingt (approved):

PR #570, opened 2022-08-24
🐛 [Portfolio] Fix OG image not display for non-Likers

nwingt:


williamchong (approved):
Maybe ellipsis the wallet for og display?

AuroraHuang22 (approved):

PR #577, opened 2022-08-29
✨ Fetch metadata by `uri`, determine and sort classes by whether it's Writing NFT

WeiJun0827:


src/util/nft.js
```diff
@@ -112,3 +112,21 @@ export async function getAccountBalance(address) {
...
+export function isValidHttpUrl(string) {
+  let url;
+
+  try {
+    url = new URL(string);
+  } catch (_) {
+    return false;
+  }
+
+  return url.protocol === 'http:' || url.protocol === 'https:';
```
williamchong, line 125:
```suggestion
  try {
    const url = new URL(string);
    return url.protocol === 'http:' || url.protocol === 'https:';
  } catch (_) {
    // no op
  }
  return false;

```

src/pages/_id/index.vue
```diff
@@ -245,6 +248,19 @@ export default {
     userAvatar() {
       return this.userInfo && this.userInfo.avatar;
     },
+    sortedOwnedNFTClassId() {
+      const sorted = [...this.ownedNFTClassId];
+      sorted.sort((a, b) => {
```
williamchong, line 253: Merge into 1 line? The first line's array is not sorted and is named so now
```suggestion
      const sorted = [...this.ownedNFTClassId].sort((a, b) => {
```

src/pages/_id/index.vue
```diff
@@ -245,6 +248,19 @@ export default {
...
+        const aIsWritingNFT = isWritingNFT(this.getNFTClassMetadataById(a))
+          ? 1
+          : 0;
+        const bIsWritingNFT = isWritingNFT(this.getNFTClassMetadataById(b))
+          ? 1
+          : 0;
```
williamchong, line 259: I think `Number(true) === 1` and `Number(false) === 0` so maybe just use `Number()`

src/pages/_id/index.vue
```diff
@@ -245,6 +248,19 @@ export default {
...
+        return bIsWritingNFT - aIsWritingNFT;
```
williamchong, line 260: Would be even clearer if we can just return 0, 1, -1 according to a/b's true false value

src/store/modules/actions/staticData.js
```diff
@@ -38,20 +38,26 @@ export async function fetchNFTPurchaseInfo({ commit }, classId) {
...
+  } = chainMetadata || {};
+  metadata = {
+    name,
+    description,
+    metadata: classMetadata,
+    parent,
+  };
+  if (isValidHttpUrl(uri)) {
+    const apiMetadata = await this.$api
+      .$get(uri)
```
williamchong, line 56: This differ from our current behaviour (which is not ideal) of always hitting nft api, need to test if this causes any unexpected issue on prod

williamchong (changes requested):

src/pages/_id/index.vue
```diff
@@ -245,6 +248,19 @@ export default {
     userAvatar() {
       return this.userInfo && this.userInfo.avatar;
     },
+    sortedOwnedNFTClassId() {
```
nwingt, line 251: I think can move this sort func to util, then my dashboard page can reuse later

src/store/modules/actions/staticData.js
WeiJun0827: I see. And I did check that [the genesis Writing NFT](https://mainnet-node.like.co/cosmos/nft/v1beta1/classes/likenft1jvywpl0f0zndhm6h8sp44esaj88mrkew226wv30ymeq5pnvux4tq760qd7) on prod has the `uri` field.

src/pages/_id/index.vue
WeiJun0827: How about migrating this function to `/mixin/portfolio` in the `my-dashboard` when both branches are merged? Since we need to call `getNFTClassMetadataById()` to get the metadata, it's a bit inconvenient to do so in util (have to `import { store }` and call `store.getter.getNFTClassMetadataById`?).

nwingt: How about a getter function in store?

WeiJun0827: Can do, thanks!!

WeiJun0827:
Currently, both `ItemCard` and `Item` components have a  twin-component `NonWritingItemCard` and `NonWritingItem` to display non-Writing NFT. The `nftClassIds` is explicitly split into two lists on the portfolio page. 
Not quite sure if we should leave the `isWritingNFT` determination outside of `ItemCard` and `Item` components, or put it inside and let components dynamically decide how to display the UI.

`// TODO: Fix the waterfall style to sink non-Writing NFT down on the portfolio page`

截圖 2022-08-31 下午10 19 45

截圖 2022-08-31 下午10 56 15

src/store/modules/getters/staticData.js
```diff
@@ -22,3 +24,14 @@ export const getNFTClassMintedCount = state => id =>
...
     0
   );
+
+export const getNFTClassIdSorter = (_, getters) => classIds => {
+  const sorted = [...classIds].sort((a, b) => {
+    const aIsWritingNFT = isWritingNFT(getters.getNFTClassMetadataById(a));
+    const bIsWritingNFT = isWritingNFT(getters.getNFTClassMetadataById(b));
+    if (aIsWritingNFT && !bIsWritingNFT) return -1;
+    if (!aIsWritingNFT && bIsWritingNFT) return 1;
+    return 0;
```
nwingt, line 34: Should sort by created time if return 0?

src/pages/_id/index.vue
```diff
@@ -249,14 +249,7 @@ export default {
...
     sortedOwnedNFTClassId() {
-      const sorted = [...this.ownedNFTClassId].sort((a, b) => {
-        const aIsWritingNFT = isWritingNFT(this.getNFTClassMetadataById(a));
-        const bIsWritingNFT = isWritingNFT(this.getNFTClassMetadataById(b));
-        if (aIsWritingNFT && !bIsWritingNFT) return -1;
-        if (!aIsWritingNFT && bIsWritingNFT) return 1;
-        return 0;
-      });
-      return sorted;
+      return this.getNFTClassIdSorter(this.ownedNFTClassId);
```
nwingt: ownedNFTClassId -> ownedNFTClassIds?

src/components/NFTPortfolio/NonWritingItem.vue
```diff
@@ -0,0 +1,96 @@
+<template>
```
nwingt, line 1: What is the significant difference between non-writing items& writing nft item? Seem very similar in layout, could be refactored

src/components/NFTPage/EventList/Table.vue
```diff
@@ -26,24 +26,27 @@
               <IconTransferMini />
             </template>
           </Label>
+          <Label v-else-if="event.event === 'mint_nft'" class="break-all" :text="$t('nft_details_page_activity_list_event_mint_nft')" />
+          <Label v-else-if="event.event === 'new_class'" class="break-all" :text="$t('nft_details_page_activity_list_event_create_class')" />
```
nwingt, line 30: Please use multiline props

src/components/NFTPage/NonWritingItemCard.vue
```diff
@@ -0,0 +1,125 @@
+<template>
```
nwingt, line 1: Again, can refactor with Writing NFT Item Card?

nwingt (commented):
I think it is better name `nonWritingNFT` as `primitiveNFT`

nwingt:
For 🔨 Gitmoji, it is mainly used for development scripts, e.g. deploy, lining...etc
Ref: https://github.com/carloscuesta/gitmoji/issues/564#issuecomment-694207677

Try using 🚧/🎨

src/store/modules/getters/staticData.js
WeiJun0827: Temporary sort by price for now.

src/components/NFTPortfolio/NonWritingItem.vue
WeiJun0827: Refactored

src/components/NFTPage/NonWritingItemCard.vue
WeiJun0827: Refactored, too

src/components/NFTPage/ItemCard.vue
```diff
@@ -155,6 +166,10 @@ export default {
       type: String,
       default: undefined,
     },
+    isWritingNFT: {
```
nwingt, line 169: So in HTML template can use `is-writing-nft` instead of `is-writing-n-f-t`
```suggestion
    isWritingNft: {
```

You may alias it with computed props 
```js
computed: {
  isWritingNFT() {
    return this.isWritingNft;
  },
},
```

WeiJun0827: Looks better, thx!

nwingt (approved):

AuroraHuang22 (approved):

PR #586, opened 2022-09-01
👽 Respect connected wallet session

nwingt:
Depends on https://github.com/likecoin/likecoin-api-private/pull/361


src/components/CivicLikerV3/Dashboard.vue
```diff
@@ -1,39 +1,40 @@
...
+      key="loading"
+      class="flex items-center justify-center min-h-[180px]"
+    >
+      <Spinner class="mx-auto" />
+    </div>
+    <CivicLikerV3PureDashboard
+      v-else
+      key="dashboard"
+      :status="status"
+      :is-signed-in="!!getAddress"
```
williamchong, line 14: Do we have UI that tells wallet user with enough stake to register a Liker ID to activate civic liker? This would be a case previously did not exist

src/locales/en.json
```diff
@@ -139,7 +139,7 @@
   "civic_dashboard_v3_summary_stake_hint_logout": "Stake at least {amount} to become {civicLiker}",
   "civic_dashboard_v3_summary_stake_hint_unfufilled": "stake {amount} more to become {civicLiker}",
   "civic_dashboard_v3_summary_stake_hint_fufilled": "You have staked {amount}.  The more you stake, the more you reward.",
-  "civic_dashboard_v3_summary_action_button_sign_in": "Register now",
+  "civic_dashboard_v3_summary_action_button_sign_in": "Connect Wallet",
```
WeiJun0827, line 142: Should we also change the field name?

src/locales/zh-Hant.json
```diff
@@ -139,7 +139,7 @@
   "civic_dashboard_v3_summary_stake_hint_logout": "成為{civicLiker}：最少委託 {amount} ",
   "civic_dashboard_v3_summary_stake_hint_unfufilled": "還需 {amount} 便能成為{civicLiker}",
   "civic_dashboard_v3_summary_stake_hint_fufilled": "已委託 {amount}。委託愈多，回饋愈多！",
-  "civic_dashboard_v3_summary_action_button_sign_in": "立即註冊",
+  "civic_dashboard_v3_summary_action_button_sign_in": "連接錢包",
```
WeiJun0827, line 142: Should we also change the field name?

williamchong (approved):

PR #592, opened 2022-09-05
🐛 Fix fetching ownCount when not login

AuroraHuang22:


WeiJun0827 (approved):

williamchong (changes requested):
i think owner info loading should not be true, nor should owner info api be called if address is not set...

AuroraHuang22:
> i think owner info loading should not be true, nor should owner info api be called if address is not set...

`isOwnerInfoLoading` should has been injected from the [mixin](https://github.com/likecoin/liker-land/blob/master/src/mixins/nft.js#:~:text=isOwnerInfoLoading%3A%20false%2C), 
and the relative API won't be called if the address is empty.
https://github.com/likecoin/liker-land/blob/master/src/mixins/nft.js#:~:text=async-,updateUserCollectedCount,-(classId%2C

williamchong:
> > i think owner info loading should not be true, nor should owner info api be called if address is not set...
> 
> `isOwnerInfoLoading` should has been injected from the [mixin](https://github.com/likecoin/liker-land/blob/master/src/mixins/nft.js#:~:text=isOwnerInfoLoading%3A%20false%2C), and the relative API won't be called if the address is empty. https://github.com/likecoin/liker-land/blob/master/src/mixins/nft.js#:~:text=async-,updateUserCollectedCount,-(classId%2C

then do we still need to check for !!address in this case after removing data?

nwingt:
I think the initial value of `isOwnerInfoLoading` can be `!!getters.getAddress` and no need to check for `!!getAddress`

PR #596, opened 2022-09-06
✨ Collect NFT with Stripe

nwingt:


src/pages/nft/class/_classId/index.vue
```diff
@@ -151,7 +151,11 @@ export default {
             this.NFTImageUrl || 'https://liker.land/images/og/writing-nft.jpg',
         },
       ],
-      link: [{ rel: 'canonical', href: `${this.$route.path}` }],
+      link: [
+        { rel: 'canonical', href: `${this.$route.path}` },
+        { rel: 'preload', href: 'https://js.stripe.com/v3', as: 'script' },
+      ],
+      script: [{ src: 'https://js.stripe.com/v3', async: true }],
```
williamchong, line 158: we better inject this script whenever a fiat payment can be triggered, but I'm not sure how

src/components/EventModal/Collect/index.vue
```diff
@@ -0,0 +1,147 @@
...
+            @click="uiCollectMethodCallback"
+          />
+        </li>
+        <li>
+          <EventModalCollectMethodButton
+            :title="$t('nft_collect_modal_method_stripe')"
+            :description="$t('nft_collect_modal_method_stripe_description')"
+            type="stripe"
+            :price="formattedFiatPrice"
+            @click="uiCollectMethodCallback"
```
williamchong, line 53: i think uiCollectMethodCallback is a strange name since it seems more like a trigger/handler function
Also strange to have ui logic depend on exist of this function or not

williamchong (approved):

PR #599, opened 2022-09-07
♻️ Use click methods for logTracker

Tsai-Hsueh-Kuan:
♻️ Use click methods for logTracker
Want to confirm that the Legacy event is mentioned here, do you think there is no need to add an event?


williamchong (changes requested):
i think we can keep `to` and use @click.native instead of click

nwingt:
@Tsai-Hsueh-Kuan Please try @williamchong suggestion

williamchong (approved):

PR #600, opened 2022-09-07
🐛 Do not fallback when history is found but empty

WeiJun0827:


src/mixins/nft.js
```diff
@@ -200,9 +201,10 @@ export default {
         } catch (error) {
           // eslint-disable-next-line no-console
           console.error(error);
+          fallbackToPrimitiveNFT = true;
         }
       }
-      if (!this.NFTHistory?.length) {
+      if (fallbackToPrimitiveNFT) {
```
williamchong, line 207: `if (!this.NFTHistory || fallbackToPrimitiveNFT)`

williamchong (changes requested):

williamchong (approved):

PR #612, opened 2022-09-13
📈 Add event sell button

Tsai-Hsueh-Kuan:
📈 Add event sell button

### DIFF
--- a/src/components/NFTPage/PriceSection.vue
+++ b/src/components/NFTPage/PriceSection.vue
@@ -72,13 +72,17 @@
           class="ml-[12px]"
           :text="$t('nft_details_page_button_sell')"
           preset="tertiary"
-          :is-disabled="true"
+          :is-disabled="false"
+          @mouseenter="handleMouseEnterSell"
+          @click.native="handleClickSell"
         />
       </ToolTips>
     </div>
   </CardV2>
 </template>
 <script>
+import { logTrackerEvent } from '~/util/EventLogger';
+
 export default {
   name: 'PriceSection',
   props: {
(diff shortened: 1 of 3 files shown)


src/components/NFTPage/PriceSection.vue
```diff
@@ -112,6 +121,27 @@ export default {
...
+        logTrackerEvent(
+          this,
+          'NFTSellButton',
+          'NFTSellButtonHandleMouseEnterSell',
+          '',
+          1
+        );
+        this.isSellHoverLogSend = true;
+      }
+    },
```
nwingt, line 144: Suggest to emit `'click-sell'` and `'hover-sell'` event for details page to log the event

src/components/NFTPage/PriceSection.vue
```diff
@@ -103,6 +107,11 @@ export default {
       default: false,
     },
   },
+  data() {
+    return {
+      isSellHoverLogSend: false,
```
nwingt, line 112:
```suggestion
      hasHoverSellButton: false,
```

nwingt (changes requested):

src/components/NFTPage/PriceSection.vue
```diff
@@ -112,6 +119,15 @@ export default {
     handleClickCollect() {
       this.$emit('collect');
     },
+    handleClickSell() {
+      this.$emit('click-sell');
+    },
+    handleMouseEnterSell() {
+      if (!this.isSellHoverLogSend) {
+        this.$emit('hover-sell');
+        this.isSellHoverLogSend = true;
```
nwingt, line 128:
```suggestion
      if (!this.hasHoverSellButton) {
        this.$emit('hover-sell');
        this.hasHoverSellButton = true;
```

nwingt (changes requested):

nwingt (approved):

src/components/NFTPage/PriceSection.vue
```diff
@@ -103,6 +105,11 @@ export default {
       default: false,
     },
   },
+  data() {
+    return {
+      hasHoverSellButton: false,
```
williamchong, line 110: Better set this in page where we send the actual log, instead of component?
Hover event should always emit, but we only want to log once

williamchong (changes requested):

williamchong (approved):

PR #619, opened 2022-09-14
✨ Split sign broadcast in collect

Tsai-Hsueh-Kuan:
✨ Split sign broadcast in collect

### DIFF
--- a/src/mixins/nft.js
+++ b/src/mixins/nft.js
@@ -16,7 +16,8 @@ import { logTrackerEvent } from '~/util/EventLogger';
 import {
   getAccountBalance,
   transferNFT,
-  sendGrant,
+  signGrant,
+  broadcastTx,
   getNFTCountByClassId,
   getISCNRecord,
   isWritingNFT,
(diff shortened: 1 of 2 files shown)


src/util/nft.js
```diff
@@ -74,12 +84,20 @@ export async function sendGrant({ senderAddress, amountInLIKE, signer }) {
...
-    expirationInMs
+    expirationInMs,
+    { broadcast: false }
   );
+  return signData;
+}
+
+export async function broadcastTx(signData, signer) {
+  const txBytes = TxRaw.encode(signData).finish();
+  const senderClient = await getSenderClientBySigner(signer);
```
nwingt, line 99: I think this can be replaced by
https://github.com/likecoin/liker-land/blob/f9fad1b2c0d400ee7b1638c28c1b5142fec1a461/src/util/nft.js#L23-L27

```suggestion
  const client = await createNFTSigningClient(signer);
  const senderClient = client.getSigningStargateClient();
```

nwingt (commented):
How about Transfer NFT?

src/util/nft.js
```diff
@@ -74,12 +74,21 @@ export async function sendGrant({ senderAddress, amountInLIKE, signer }) {
...
+    { broadcast: false }
   );
+  return signData;
+}
+
+export async function broadcastTx(signData, signer) {
+  const txBytes = TxRaw.encode(signData).finish();
+  const client = await createNFTSigningClient(signer);
+  const senderClient = client.getSigningStargateClient();
+  const { transactionHash } = await senderClient.broadcastTx(txBytes);
```
nwingt, line 91:
```suggestion
  const client = await createNFTSigningClient(signer);
  const senderClient = client.getSigningStargateClient();
  const txBytes = TxRaw.encode(signData).finish();
  const { transactionHash } = await senderClient.broadcastTx(txBytes);
```

nwingt (changes requested):

src/util/nft.js
Tsai-Hsueh-Kuan: Is this only changed in order?

src/mixins/nft.js
```diff
@@ -455,11 +464,20 @@ export default {
...
+          'NFTTransferSignApproved',
           this.classId,
           1
         );
         this.uiSetTxStatus(TX_STATUS.PROCESSING);
+        const txHash = await broadcastTx(signData, this.getSigner);
+        logTrackerEvent(
+          this,
+          'NFT',
+          'NFTCollectBroadcastTxComplete',
```
williamchong, line 476:
```suggestion
          'NFTTransferBroadcastTxComplete',
```

williamchong (changes requested):

williamchong (approved):

PR #621, opened 2022-09-14
✨ [CIvic] Add Writing NFT campaign CTA in dashboard

nwingt:


src/components/CivicLikerV3/PureDashboard.vue
```diff
@@ -169,22 +169,32 @@
...
+          class="mt-32"
           path="civic_dashboard_v3_intro_description"
           tag="p"
         />
+
+        <div class="flex justify-center mt-[32px]">
+          <ButtonV2
+            :text="$t('about_nft_page_title')"
+            preset="secondary"
+            to="/campaign/writing-nft"
```
williamchong, line 194: Should we prefer named route?

nwingt: I prefer simply path when no params/ query

williamchong: hard to find and replace

williamchong (approved):

PR #622, opened 2022-09-14
✨ Merge NFT events in DB into on-chain events

WeiJun0827:
Take [the genesis NFT](https://liker.land/nft/class/likenft1yhsps5l8tmeuy9y7k0rjpx97cl67cjkjnzkycecw5xrvjjp6c5yqz0ttmc) for example:
before:
截圖 2022-09-14 下午5 58 27

after:
截圖 2022-09-14 下午5 58 05

### DIFF
--- a/src/mixins/nft.js
+++ b/src/mixins/nft.js
@@ -261,24 +261,7 @@ export default {
     },
     async updateNFTHistory() {
       this.isHistoryInfoLoading = true;
-      let data;
-      let nextKey;
-      let count;
-      let events = [];
-      do {
-        // eslint-disable-next-line no-await-in-loop
-        ({ data } = await this.$api.get(
-          getNFTEvents({
-            classId: this.classId,
-            key: nextKey,
-            limit: NFT_INDEXER_LIMIT_MAX,
-          })
-        ));
-        nextKey = data.pagination.next_key;
-        ({ count } = data.pagination);
-        events = events.concat(data.events);
-      } while (count === NFT_INDEXER_LIMIT_MAX);
-      const historyOnChain = formatNFTEventsToHistory(events);
+      const historyOnChain = await this.getNFTEventsAll();
       this.NFTHistory = historyOnChain;
 
       if (this.isWritingNFT) {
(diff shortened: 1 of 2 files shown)


src/mixins/nft.js
```diff
@@ -256,39 +261,34 @@ export default {
     },
     async updateNFTHistory() {
       this.isHistoryInfoLoading = true;
-      let hasHistoryInDatabase = false;
+      const historyOnChain = await this.getNFTEventsAll();
+      this.NFTHistory = historyOnChain;
+
       if (this.isWritingNFT) {
```
nwingt, line 267: The history UI might flash twice if `isWritingNFT` is `true`, since `this.NFTHistory` is assigned here and assigned again later after `isWritingNFT` check

Should assign `this.NFTHistory` after `isWritingNFT` check

src/mixins/nft.js
```diff
@@ -256,39 +261,34 @@ export default {
...
           );
-          this.NFTHistory = data.list;
-          hasHistoryInDatabase = true;
+          const historyInDB = data.list;
+          const txhashToEventMap = new Map();
+          historyOnChain
+            .filter(
+              e =>
+                e.event !== 'new_class' &&
+                e.event !== 'mint_nft' &&
```
nwingt, line 278:
```suggestion
                !['new_class', 'mint_nft'].includes(e.event)  &&
```

src/mixins/nft.js
```diff
@@ -297,6 +297,26 @@ export default {
...
+        ({ data } = await this.$api.get(
+          getNFTEvents({
+            classId: this.classId,
+            key: nextKey,
+            limit: NFT_INDEXER_LIMIT_MAX,
+          })
+        ));
+        nextKey = data.pagination.next_key;
+        ({ count } = data.pagination);
+        events = events.concat(data.events);
```
nwingt, line 316: `events` can be a `const`
```suggestion
        events.push(...data.events);
```

src/mixins/nft.js
```diff
@@ -256,39 +261,33 @@ export default {
     },
     async updateNFTHistory() {
       this.isHistoryInfoLoading = true;
-      let hasHistoryInDatabase = false;
+      const historyOnChain = await this.getNFTEventsAll();
+      let history = historyOnChain;
```
williamchong, line 265: Not sure if you want to clone this in case some one modify `history` but accidentally modify `historyOnChain`

src/mixins/nft.js
```diff
@@ -256,39 +261,33 @@ export default {
...
+          const eventMap = new Map();
+          historyOnChain
+            .filter(
+              e =>
+                !['new_class', 'mint_nft'].includes(e.event) &&
+                e.toWallet !== LIKECOIN_NFT_API_WALLET
+            )
+            .forEach(e => eventMap.set(`${e.txHash}-${e.nftId}`, e));
+          // overwrite on-chain event with db event
+          historyInDB.forEach(e => eventMap.set(`${e.txHash}-${e.nftId}`, e));
```
williamchong, line 282: I think we only need price info, can we only inject price info for an hash-id that already exists?

src/mixins/nft.js
WeiJun0827: `history` and `historyOnChain` only live in `updateNFTHistory()` method. If we assign `history` to `this.NFTHistory` (`this.NFTHistory = history;`) after all the process, can anyone still be able to modify it from somewhere outside the method?

williamchong: if it only live in this function i guess it is ok, but later if this function expands maybe someone will mess this up

williamchong (approved):

PR #629, opened 2022-09-16
📈 Add collect event add_to_cart

Tsai-Hsueh-Kuan:
📈 Add collect event add_to_cart

### DIFF
--- a/src/mixins/nft.js
+++ b/src/mixins/nft.js
@@ -364,6 +364,12 @@ export default {
         await this.initIfNecessary();
         this.fetchUserCollectedCount();
         this.userAccountBalanceFetch = getAccountBalance(this.getAddress);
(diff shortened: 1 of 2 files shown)


src/mixins/nft.js
```diff
@@ -364,6 +364,12 @@ export default {
         await this.initIfNecessary();
         this.fetchUserCollectedCount();
         this.userAccountBalanceFetch = getAccountBalance(this.getAddress);
+        logPurchaseFlowEvent(this, 'add_to_cart', {
+          txHash,
```
williamchong, line 368: these is no `txHash` yet.
Also what does `initIfNecessary` do, should we put the log before it?

williamchong (changes requested):

PR #631, opened 2022-09-16
🚸 Do not block detail page when loading

WeiJun0827:


src/pages/nft/class/_classId/index.vue
```diff
@@ -201,15 +201,15 @@ export default {
...
-      await Promise.all(promises);
+      this.updateDisplayNameList(this.iscnOwner);
+      this.updateNFTOwners();
+      this.updateNFTHistory();
+      this.getLIKEPrice();
+      const blockingPromises = [this.fetchISCNMetadata()];
+      if (this.isWritingNFT) {
+        blockingPromises.push(this.updateNFTPurchaseInfo());
+      }
+      await Promise.all(blockingPromises);
```
williamchong, line 212: Should we add catch for each of the promises if the error is not critical? If we use one catch or not awaiting, there might be unhandled promise that would kill the whole function

williamchong (changes requested):

PR #632, opened 2022-09-16
🎨 Move balance to vuex store

AuroraHuang22:


src/store/modules/getters/wallet.js
```diff
@@ -3,3 +3,4 @@ export const getSigner = state => state.signer;
 export const getConnector = state => state.connector;
 export const getLikerInfo = state => state.likerInfo;
 export const walletMethodType = state => state.methodType;
+export const walletLIKEBalance = state => Number(state.likeBalance).toFixed(2);
```
nwingt, line 6: Why `toFixed` here?  Should handle rounding & formatting at the very end with `formatNumber()`

src/store/modules/actions/wallet.js
```diff
@@ -79,3 +81,12 @@ export async function initIfNecessary({ dispatch }) {
     await dispatch('initWallet', { accounts, offlineSigner, method });
   }
 }
+
+export async function walletFetchLIKEBalance({ commit }, address) {
```
nwingt, line 85: This action doesn't need `address` as param when we can get it from the state.

```suggestion
export async function walletFetchLIKEBalance({ commit, state }) {
  const { address } = state;
```

src/store/modules/actions/wallet.js
```diff
@@ -79,3 +81,12 @@ export async function initIfNecessary({ dispatch }) {
...
+  const c = await getNFTQueryClient();
+  const client = await c.getQueryClient();
+  const balance = amountToLIKE(
+    await client.bank.balance(address, LIKECOIN_CHAIN_MIN_DENOM)
+  );
```
nwingt, line 90: I think this part can be refactored to util
There will be some scenarios for getting the balance of an address

src/util/nft.js
```diff
@@ -110,24 +110,6 @@ export async function signTransferNFT({
...
-}
-
-export async function getAccountBalance(address) {
-  const c = await getNFTQueryClient();
-  const client = await c.getQueryClient();
-  return amountToLIKE(
-    await client.bank.balance(address, LIKECOIN_CHAIN_MIN_DENOM)
-  );
-}
-
```
nwingt, line 130: I think can keep these functions here

src/store/modules/actions/wallet.js
```diff
@@ -79,3 +80,9 @@ export async function initIfNecessary({ dispatch }) {
     await dispatch('initWallet', { accounts, offlineSigner, method });
   }
 }
+
+export async function walletFetchLIKEBalance({ commit, state }) {
+  const { address } = state;
+  const balance = await fetchLIKEBalance(address);
+  commit(types.WALLET_LIKE_BALANCE, balance);
+}
```
nwingt, line 88:
```suggestion
export async function walletFetchLIKEBalance({ commit, state }) {
  const { address } = state;
  const balance = await getAccountBalance(address);
  commit(types.WALLET_LIKE_BALANCE, balance);
}
```

src/store/modules/actions/wallet.js
```diff
@@ -1,8 +1,9 @@
 import { LikeCoinWalletConnector } from '@likecoin/wallet-connector';
 
 import { LIKECOIN_WALLET_CONNECTOR_CONFIG } from '@/constant/network';
-import { getAddressLikerIdMinApi } from '~/util/api';
 import * as types from '@/store/mutation-types';
+import { fetchLIKEBalance } from '~/util/wallet';
```
nwingt, line 5:
```suggestion
import { getAccountBalance } from '~/util/nft';
```

nwingt (changes requested):
Seems keeping `amountToLIKE` and `getAccountBalance` in `util/nft` for now is more simple, let us rethink the files structure later

nwingt (approved):

PR #639, opened 2022-09-19
⚡️ Cache userInfo in staticData store

WeiJun0827:
- Move `iscnOwnerInfo` from `data` to `computed` in nft mixin
- Move `displayNameList`, `avatarList`, and `civicLikerList` to staticData store
- Add `userInfoLastQueryTimestamps` in staticData store to track TTL of userInfo cache
- Support `fetchUserInfoByAddress()` and `lazyGetUserInfoByAddress()` actions
- Replace most `$api.get(getUserMinByAddress(address))` with above actions
- Rename `getAddressLikerIdMinApi()` to `getUserMinByAddress()`
- No longer query all owner's userInfo, since we only need the owner count on the portfolio page, campaign page, and share page. Other functions such as `updateNFTClassMetadata()`, `updateNFTHistory()`, and `mounted()` already call the necessary `updateDisplayNameList()` to display sufficient userInfo.


src/store/modules/actions/staticData.js
```diff
@@ -15,18 +17,48 @@ export async function fetchUserInfo({ commit, state }, opts) {
...
+  commit(TYPES.STATIC_SET_USER_INFO_BY_ADDRESS, { address, userInfo });
+  commit(TYPES.STATIC_SET_USER_INFO_LAST_QUERY_TIMESTAMP, {
+    address,
+    timestamp: Date.now(),
+  });
+  try {
+    userInfo = await this.$api.$get(api.getUserMinByAddress(address));
+    commit(TYPES.STATIC_SET_USER_INFO_BY_ADDRESS, { address, userInfo });
+    const { user: id } = userInfo;
+    if (id) commit(TYPES.STATIC_SET_USER_INFO_BY_ID, { id, userInfo });
```
nwingt, line 48: Do we have to store user info with ID as key? 
I think we can use a getter to return `userInfosById` by reducing `userInfosByAddress`.

src/store/modules/actions/staticData.js
```diff
@@ -15,18 +17,48 @@ export async function fetchUserInfo({ commit, state }, opts) {
...
+    avatar: `https://avatars.dicebear.com/api/identicon/${address}.svg`,
+  };
+  commit(TYPES.STATIC_SET_USER_INFO_BY_ADDRESS, { address, userInfo });
+  commit(TYPES.STATIC_SET_USER_INFO_LAST_QUERY_TIMESTAMP, {
+    address,
+    timestamp: Date.now(),
+  });
+  try {
+    userInfo = await this.$api.$get(api.getUserMinByAddress(address));
+    commit(TYPES.STATIC_SET_USER_INFO_BY_ADDRESS, { address, userInfo });
```
nwingt, line 46: I think it is better to commit once at the end of this action for better performance, so the UI would not re-render multiple times

src/util/api/index.js
```diff
@@ -97,7 +97,7 @@ export const getUserMinAPI = (id, { types = [] } = {}) =>
   `${LIKECOIN_API_BASE}/users/id/${id}/min?${querystring.stringify({
     type: types.join(','),
   })}`;
-export const getAddressLikerIdMinApi = addr =>
+export const getUserMinByAddress = addr =>
```
nwingt, line 100:
```suggestion
export const getUserInfoMinByAddress = addr =>
```

src/store/modules/staticData.js
```diff
@@ -14,7 +16,9 @@ import * as getters from './getters/staticData';
 import * as actions from './actions/staticData';
 
 const state = () => ({
-  userInfos: {},
+  userInfosById: {},
+  userInfosByAddress: {},
```
nwingt, line 20:
```suggestion
  userInfoMapByAddress: {},
```

src/store/modules/staticData.js
```diff
@@ -14,7 +16,9 @@ import * as getters from './getters/staticData';
...
+  userInfoLastQueryTimestamps: {},
```
nwingt, line 21:
```suggestion
  userInfoLastQueryTimestampMap: {},
```

src/store/modules/getters/staticData.js
```diff
@@ -1,6 +1,12 @@
 import { isWritingNFT } from '~/util/nft';
 
-export const getUserInfoById = state => id => state.userInfos[id];
+export const getUserInfoById = state => id => state.userInfosById[id];
+
```
nwingt, line 4:
```suggestion
export const userInfosById = state => Object.values(state.userInfosByAddress).reduce(/* code */);

export const getUserInfoById = state => id => state.userInfosById[id];

```

WeiJun0827: Would we still need to maintain id-userInfo map? Since it's going to be removed soon.
By the way, will this modification make the time complexity of `getUserInfoById` become O(N) if we use `reduce` here?

nwingt: for data structure, maybe replace the `userInfosById` with a map with userId as key and address as value is better, so you can leverage the `userInfoByAddressMap`

WeiJun0827: I see, it looks better. But should we further modify `getUserInfoById()` related logic or keep it isolated in this PR?

src/pages/nft/class/_classId/share/index.vue
```diff
@@ -331,12 +333,7 @@ export default {
     },
     async updateReferrerInfo() {
       try {
-        const referrerInfo = await this.$api.get(
-          getAddressLikerIdMinApi(this.referrer),
-          {
-            validateStatus: code => code < 500 && code !== 400,
-          }
-        );
+        const referrerInfo = await this.fetchUserInfoById(this.referrer);
```
nwingt, line 336: Isn't `referrer` a wallet address?

src/mixins/nft.js
```diff
@@ -284,11 +287,8 @@ export default {
         console.error(error);
       }
     },
-    async updateNFTOwners() {
-      await this.fetchNFTOwners(this.classId);
-      this.updateDisplayNameList(
-        Object.keys(this.getNFTClassOwnerInfoById(this.classId))
-      );
+    updateNFTOwners() {
+      this.fetchNFTOwners(this.classId);
```
nwingt, line 291:
```suggestion
      return this.fetchNFTOwners(this.classId);
```

src/store/modules/staticData.js
```diff
@@ -14,20 +17,32 @@ import * as getters from './getters/staticData';
 import * as actions from './actions/staticData';
 
 const state = () => ({
-  userInfos: {},
+  userInfosById: {},
```
nwingt, line 20: Should remain unchanged for this PR?
```suggestion
  userInfos: {},
```

src/store/modules/actions/staticData.js
```diff
@@ -29,6 +31,43 @@ export async function fetchUserInfo({ commit, state }, opts) {
...
+    avatar: `https://avatars.dicebear.com/api/identicon/${address}.svg`,
+  };
+
+  let promise = state.fetching.userInfo[address];
+  if (promise) {
+    userInfo = await promise;
+    return userInfo;
+  }
+
+  try {
```
nwingt, line 46: `promise` might throw error
```suggestion
  try {
    let promise = state.fetching.userInfo[address];
    if (promise) {
      userInfo = await promise;
      return userInfo;
    }
```

src/store/modules/getters/staticData.js
```diff
@@ -1,6 +1,9 @@
 import { isWritingNFT } from '~/util/nft';
 
-export const getUserInfoById = state => id => state.userInfos[id];
+export const getUserInfoById = state => id => state.userInfosById[id];
```
nwingt, line 3: Should remain unchanged for this PR?
```suggestion
export const getUserInfoById = state => id => state.userInfos[id];
```

src/store/modules/staticData.js
WeiJun0827: right, thanks

nwingt (approved):

PR #655, opened 2022-09-22
🚸 Use popup for settings

nwingt:
Requires https://github.com/likecoin/like-co/pull/1665

### DIFF
--- a/src/util/links.js
+++ b/src/util/links.js
@@ -2,4 +2,4 @@ import { LIKE_CO_URL_BASE } from '~/constant';
 
 export const getLikeCoURL = (path = '') => `${LIKE_CO_URL_BASE}${path}`;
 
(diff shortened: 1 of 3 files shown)


src/util/links.js
```diff
@@ -2,4 +2,4 @@ import { LIKE_CO_URL_BASE } from '~/constant';
 
 export const getLikeCoURL = (path = '') => `${LIKE_CO_URL_BASE}${path}`;
 
-export const getLikerIdSettingsURL = () => getLikeCoURL('/in/settings');
+export const getLikerIdSettingsURL = () => getLikeCoURL('/in/settings?popup');
```
williamchong, line 5: set `popup=1` to be more explicit?

williamchong (approved):

PR #665, opened 2022-09-27
✨ Use Web Share API for sharing URL

Tsai-Hsueh-Kuan:
🚧 Wait check title/text

### DIFF
--- a/src/mixins/clipboard.js
+++ b/src/mixins/clipboard.js
@@ -16,7 +16,7 @@ export default {
         url,
       });
     },
-    shareURLPath({ title, text, path }, options) {
+    shareURLPath({ title, text, path, ...options }) {
       const host = `${window.location.protocol}//${window.location.host}`;
       const url = `${host}${path}`;
       if (navigator.share) {
--- a/src/mixins/clipboard.js
+++ b/src/mixins/clipboard.js
@@ -6,8 +6,16 @@ export default {
   mixins: [alertMixin],
   methods: {
     copyToClipboard(text, { alertMessage = this.$t('clipboard_copied') } = {}) {
-      copyToClipboard(text);
-      this.alertPromptSuccess(alertMessage);
+      if (navigator.share) {
+        navigator.share({
+          title: 'Just title',
+          text: 'Just text',
+          url: text,
+        });
+      } else {
+        copyToClipboard(text);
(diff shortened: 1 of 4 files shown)


src/mixins/clipboard.js
```diff
@@ -6,8 +6,16 @@ export default {
   mixins: [alertMixin],
   methods: {
     copyToClipboard(text, { alertMessage = this.$t('clipboard_copied') } = {}) {
-      copyToClipboard(text);
-      this.alertPromptSuccess(alertMessage);
+      if (navigator.share) {
+        navigator.share({
+          title: 'Just title',
+          text: 'Just text',
```
nwingt, line 12: -`copyToClipboard` is a generic method, I think we should check the text is a URL first, it might be plain text
- Should pass `title`/`text` from function params, not hardcoded `Just title` or `Just text`

Tsai-Hsueh-Kuan: but i check `copyToClipboard` only use here
`  copyURLPath(path, options) {
      const host = `${window.location.protocol}//${window.location.host}`;
      const url = `${host}${path}`;
      this.copyToClipboard(url, options);
    },`
so input text is a URL... Or worry about using it elsewhere in the future？

nwingt: Yes, this function might use elsewhere in the future

src/mixins/clipboard.js
```diff
@@ -9,10 +9,21 @@ export default {
...
     },
-    copyURLPath(path, options) {
+    share({ title, text, url }) {
+      navigator.share({
+        title,
+        text,
+        url,
+      });
+    },
+    shareURLPath({ title, text, path }, options) {
```
nwingt, line 19: Maybe just merge into one object
```suggestion
    shareURLPath({ title, text, path, ...options }) {
```

Tsai-Hsueh-Kuan: But this should not use options but use alertMessage? 
`      this.shareURLPath({
        title: this.userDisplayName,
        text: this.userDisplayName,
        path: `/${wallet}?referrer=${referrer}`,
        alertMessage: this.$t('tooltip_share_done'),
      });`

nwingt (approved):

PR #679, opened 2022-09-29
⚡️ Lazy load iscn and wallet library

williamchong:
The code split avoids loading @cosmjs thus the huge `libsodium` in front page for new users
Thus removing a few seconds js parsing block on main thread

Pre:
image

Post:
image

### DIFF
--- a/src/store/modules/actions/staticData.js
+++ b/src/store/modules/actions/staticData.js
@@ -1,10 +1,7 @@
 import * as TYPES from '@/store/mutation-types';
 import * as api from '@/util/api';
-import {
-  getClassInfo,
-  isValidHttpUrl,
-  formatOwnerInfoFromChain,
-} from '@/util/nft';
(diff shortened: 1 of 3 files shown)


src/store/modules/actions/staticData.js
```diff
@@ -88,7 +85,10 @@ export async function lazyGetNFTPurchaseInfo({ getters, dispatch }, classId) {
 
 export async function fetchNFTMetadata({ commit }, classId) {
   let metadata;
-  const chainMetadata = await getClassInfo(classId);
+  // const chainMetadata = await getClassInfo(classId);
```
nwingt, line 88: Should remove this?

nwingt (approved):

src/store/modules/actions/staticData.js
williamchong: want to remind ppl this is a performance hack, maybe i add a comment

PR #682, opened 2022-10-03
♻️ Split NFT classId store to created classIds and collected NFTs

WeiJun0827:


williamchong (approved):
Why split though

nwingt (commented):
I suggest that we just store user collected class IDs and storing classInfos in another state, then we can obtain classInfos through a getter

nwingt:
You might consider moving `nftClassPurchaseInfo`, `nftClassMetadata` and `nftClassOwnerInfo` from `staticData` to `nft` module

PR #683, opened 2022-10-03
🐛 Fix NFT supply table

nwingt:


src/util/writing-nft.js
```diff
@@ -22,8 +22,8 @@ export function getPowerFactor(batch) {
   return (
     1 -
     (batch - PRICE_GROWTH_POWER_DEGRADE_BATCH_BEGIN) /
-      (PRICE_GROWTH_POWER_DEGRADE_BATCH_BEGIN -
-        PRICE_GROWTH_POWER_DEGRADE_BATCH_END)
+      (PRICE_GROWTH_POWER_DEGRADE_BATCH_END -
+        PRICE_GROWTH_POWER_DEGRADE_BATCH_BEGIN)
```
williamchong, line 26: Would this function be helpful?
https://github.com/likecoin/likecoin-api-public/blob/master/src/util/api/likernft/purchase.js#L33

nwingt: should we refactor it into a lib?

nwingt: Ping @williamchong

PR #690, opened 2022-10-06
✨ Support class sorting in portfolio page

WeiJun0827:


src/store/modules/nft.js
```diff
@@ -1,41 +1,271 @@
...
+      (acc, val) => acc + val.length,
+      0
+    ),
+  getUserLastCollectedTimestampByAddress: state => address =>
+    state.userLastCollectedTimestampMap[address],
+  getCreatedClassSorter: (_, getters) => (
+    classIds,
+    orderBy,
+    order = ORDER.DESC
+  ) => {
```
nwingt, line 77: I prefer using object for params, in case we add more in the future
```suggestion
  getCreatedClassIdSorter: (_, getters) => ({
    classIds,
    orderBy,
    order = ORDER.DESC
  }) => {
```

src/util/nft.js
```diff
@@ -11,6 +10,24 @@ import {
...
+export const ORDER_CREATED_CLASS_ID_BY = {
+  PRICE: 'PRICE',
+  ISCN_TIMESTAMP: 'ISCN_TIMESTAMP',
+};
+
+export const ORDER_COLLECTED_CLASS_ID_BY = {
+  PRICE: 'PRICE',
+  LAST_COLLECTED_NFT: 'LAST_COLLECTED_NFT',
+  NFT_OWNED_COUNT: 'NFT_OWNED_COUNT',
+};
```
nwingt, line 24: This can be merged as one object?

src/store/modules/nft.js
```diff
@@ -1,41 +1,271 @@
...
+      }
+      if (A === null) return 1;
+      if (B === null) return -1;
+      switch (order) {
+        case ORDER.ASC:
+          return A - B;
+        case ORDER.DESC:
+        default:
+          return B - A;
+      }
```
nwingt, line 141: This part of code can be refactorized?

src/store/modules/nft.js
```diff
@@ -1,41 +1,271 @@
...
+
+  getCollectedClassSorter: (_, getters) => (
+    classIds,
+    address,
+    orderBy,
+    order = ORDER.DESC
+  ) => {
+    const sorted = [...classIds].sort((a, b) => {
+      const isWritingNFTCompareResult = compareIsWritingNFT(getters, a, b);
+      if (isWritingNFTCompareResult) return isWritingNFTCompareResult;
```
nwingt, line 115: Same as above

src/store/modules/nft.js
```diff
@@ -1,41 +1,271 @@
...
+  isValidHttpUrl,
+  formatOwnerInfoFromChain,
+  formatNFTInfo,
+} from '~/util/nft';
+import * as TYPES from '../mutation-types';
 
 const state = () => ({
+  nftClassPurchaseInfo: {},
+  nftClassMetadata: {},
+  nftClassOwnerInfo: {},
```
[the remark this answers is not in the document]
WeiJun0827: Should we also rename the getter at this stage?

src/util/nft.js
WeiJun0827: Will it cause misusing in different scenarios?

src/store/modules/nft.js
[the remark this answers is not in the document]
nwingt: We can keep the getter first

src/util/nft.js
nwingt: I think it is not intuitive for introducing `created` and `collected` concept for sorting

src/store/modules/nft.js
```diff
@@ -1,41 +1,264 @@
...
 
+function compareIsWritingNFT(getters, classIdA, classIdB) {
+  const aIsWritingNFT = isWritingNFT(getters.getNFTClassMetadataById(classIdA));
+  const bIsWritingNFT = isWritingNFT(getters.getNFTClassMetadataById(classIdB));
+  if (aIsWritingNFT && !bIsWritingNFT) return -1;
+  if (!aIsWritingNFT && bIsWritingNFT) return 1;
+  return 0;
+}
+
+function guardValueAndSort(X, Y, order) {
```
nwingt, line 53:
```suggestion
function compareNumber(X, Y, order) {
```

src/util/nft.js
nwingt: Anyway, I think we keep separated for now

WeiJun0827: OK

nwingt (approved):

PR #701, opened 2022-10-13
✨ Use mintscan for tx view in nft events

williamchong:


src/constant/index.js
```diff
@@ -45,6 +45,10 @@ export const CIVIC_LIKER_V3_STAKING_ENDPOINT = IS_TESTNET
   ? 'https://likecoin-public-testnet-5.netlify.app/validators'
   : 'https://dao.like.co/validators';
 
+export const LIKECOIN_CHAIN_VIEW_TX = IS_TESTNET
+  ? 'https://node.testnet.like.co/cosmos/tx/v1beta1/txs'
+  : 'hhttps://mintscan.io/likecoin/txs';
```
nwingt, line 50:
```suggestion
  : 'https://mintscan.io/likecoin/txs';
```

nwingt (changes requested):

src/constant/index.js
williamchong: updated

nwingt (approved):

PR #702, opened 2022-10-13
✨ Add userId to sitemap

williamchong:
Temp solution: add some creator owned by api wallet to sitemap
we should be using creator and collector api in the future to obtain wallets

### DIFF
--- a/src/config/sitemap.js
+++ b/src/config/sitemap.js
@@ -26,12 +26,12 @@ async function getSitemapRoutes() {
     ...[newClassRes, topClassRes].map(r => (r.data || {}).classes || [])
   );
   const classIds = classes.map(c => c.id);
-  const uniqueClassIdRoutes = [...new Set(classIds)].map(
+  const nftDetailsPageRoutes = [...new Set(classIds)].map(
     id => `/nft/class/${id}`
   );
   const users = ((creatorRes.data || {}).creators || []).map(c => c.account);
-  const userIdRoutes = users.map(id => `/${id}`);
-  return uniqueClassIdRoutes.concat(userIdRoutes);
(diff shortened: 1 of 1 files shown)


src/config/sitemap.js
```diff
@@ -10,21 +10,28 @@ const LIKECOIN_NFT_API_WALLET = IS_TESTNET
...
+  const classes = [].concat(
+    ...[newClassRes, topClassRes].map(r => (r.data || {}).classes || [])
+  );
+  const classIds = classes.map(c => c.id);
+  const uniqueClassIdRoutes = [...new Set(classIds)].map(
+    id => `/nft/class/${id}`
+  );
+  const users = ((creatorRes.data || {}).creators || []).map(c => c.account);
+  const userIdRoutes = users.map(id => `/${id}`);
+  return uniqueClassIdRoutes.concat(userIdRoutes);
```
nwingt, line 34: I think it is better to rename the const for readibilty
```suggestion
  const routes = [...new Set(classIds)].map(
    id => `/nft/class/${id}`
  );
  const users = ((creatorRes.data || {}).creators || []).map(c => c.account);
  const portfolioPageRoutes = users.map(id => `/${id}`);
  return routes.concat(portfolioPageRoutes);
```

nwingt: btw should we make sure `portfolioPageRoutes` is unique as well?

nwingt (commented):
commit message should be `✨ Add portfolio pages to sitemap`?

src/config/sitemap.js
williamchong: `likechain/likenft/v1/creator` does not return repeated id, but we can write a dedup here for future use

nwingt (approved):

PR #706, opened 2022-10-14
Skip wallet connection for collecting NFT

nwingt:
Extend from #705


src/mixins/nft.js
```diff
@@ -392,14 +392,24 @@ export default {
...
           price: this.purchaseInfo.price,
           classId: this.classId,
-        });
-        await this.initIfNecessary();
-        this.fetchUserCollectedCount();
-        this.walletFetchLIKEBalance();
+        };
+        logPurchaseFlowEvent(this, 'add_to_cart', purchaseEventParams);
+        logPurchaseFlowEvent(this, 'begin_checkout', purchaseEventParams);
+        logPurchaseFlowEvent(this, 'add_shipping_info', purchaseEventParams);
```
williamchong, line 402: I think user without wallet has no `shipp_info` since you ship nft to their wallet

williamchong (approved):

PR #709, opened 2022-10-17
🔧 Use cosmos lcd to get NFT data

AuroraHuang22:


williamchong:
Does `key` and `limit` works with this api? or is it the new syntax `pagination.limit`?

PR #712, opened 2022-10-19
👽 [Cosmostation] Support for WC Direct Sign

nwingt:


williamchong (approved):
What does debug do now?

nwingt:
Nothing, just leave it there if we need to test sth out on production

PR #717, opened 2022-10-20
🎨 Respect dual address prefix in getNFTs function

WeiJun0827:


src/util/cosmos.js
```diff
@@ -14,3 +16,12 @@ export function convertAddressPrefix(address, prefix = 'like') {
   const { words } = bech32.decode(address);
   return bech32.encode(prefix, words);
 }
+
+export function getAllowedAddresses(address) {
```
nwingt, line 20:
```suggestion
export function deriveAllPrefixedAddresses(address) {
```

src/store/modules/nft.js
```diff
@@ -230,13 +231,19 @@ const actions = {
         nextKey = data.pagination.next_key;
         ({ count } = data.pagination);
         nfts.push(...data.nfts);
-      } while (nextKey);
+      } while (count === NFT_INDEXER_LIMIT_MAX);
       // sort by last colleted by default
       return nfts.map(formatNFTInfo).sort((a, b) => b.timestamp - a.timestamp);
```
nwingt, line 236: Is this sorting necessary? Seem should be controlled by getters

WeiJun0827: If two NFTs get the same rank in the comparator (e.g. same price), the UI will display the query result's default order(`ORDER BY id` in the table). That will be a little arbitrary for users. So I sort the NFTs here to achieve the effect similar to `ORDER BY ${rule from getter} THEN BY timestamp DESC`. 
Should we keep this and add some comments here?

WeiJun0827:
Since it's hard to maintain fall-through logic simply in a switch statement when having 3+ cases, I use an additional if statement outside of the switch. @nwingt please check if it's fine, thanks!

nwingt:
> Since it's hard to maintain fall-through logic simply in a switch statement when having 3+ cases, I use an additional if statement outside of the switch. @nwingt please check if it's fine, thanks!

@WeiJun0827 Simply not working?

```js
if (X !== Y) {
  break;
}
```

WeiJun0827:
@nwingt Works for `ORDER_CREATED_CLASS_ID_BY` that has only 2 cases. 
But for `ORDER_COLLECTED_CLASS_ID_BY`: 
`case ORDER_COLLECTED_CLASS_ID_BY.PRICE` it will first fall through `case ORDER_COLLECTED_CLASS_ID_BY.NFT_OWNED_COUNT` then goes to `default`. 
Not sure if it's the effect we want?

nwingt:
I think it is ok 👍🏻

WeiJun0827:
Cool, I've updated the code, please check, thanks!!

nwingt (approved):

williamchong (approved):

PR #720, opened 2022-10-20
📈 Convert LIKE to roughly USD in purchase flow event

williamchong:


src/util/EventLogger.js
```diff
@@ -96,15 +101,21 @@ export function logPurchaseFlowEvent(
...
-        value: price,
+        value: actualPrice,
         items: [
           {
             item_id: classId,
             item_name: name,
-            currency,
-            price,
+            currency: actualCurrency,
+            price: actualPrice,
```
nwingt, line 118: I think should use this computed prop
https://github.com/likecoin/liker-land/blob/64acab5bd2a51d9dac4222ff7720ad535e08610b/src/mixins/nft.js#L146-L148

for this
https://github.com/likecoin/liker-land/blob/64acab5bd2a51d9dac4222ff7720ad535e08610b/src/mixins/nft.js#L397

williamchong: Rewrote the last commit and force pushed

nwingt (approved):

PR #725, opened 2022-10-21
✨ Add creator's NFT mint subscription on empty portfolio

nwingt:
# New UI
Entry

## Subscription Email
email

##  Confirm subscription prompt
subscribe1
subscribe2

## Unsubscribe prompt
unsubscribe2
unsubscribe1


src/components/NFTPortfolio/EmptyForm.vue
```diff
@@ -0,0 +1,132 @@
...
+        <span class="font-[600] text-like-green" place="creator">{{ formattedCreatorDisplayName }}</span>
+      </i18n>
+    </Label>
+    <div class="flex gap-[16px]">
+      <TextField
+        ref="emailTextField"
+        :text="$t('portfolio_empty_title')"
+        :placeholder="$t('portfolio_empty_email_placeholder')"
+        :is-disabled="isLoading"
+        @input="handleEmailInput"
```
williamchong, line 20: can we add `<input>`'s `type=email` here?

williamchong (changes requested):
We probably need to figure out how to integrate sendgrid and new nft class event to here before release

src/server/api/routes/nft/mint-subscription.js
```diff
@@ -0,0 +1,102 @@
...
+    const subscriptionDoc = await nftMintSubscriptionCollection
+      .doc(subscriptionId)
+      .get();
+    if (!subscriptionDoc.exists) {
+      res.status(404).send('SUBSCRIPTION_NOT_FOUND');
+      return;
+    }
+    const { ts, ...data } = subscriptionDoc.data();
+    res.json({
+      ...data,
```
williamchong, line 84: we probably don't want to show email in a semi-public api
is it necessary for us to have a get api by id?

src/server/api/routes/nft/mint-subscription.js
```diff
@@ -0,0 +1,102 @@
...
+      ts: ts.toMillis(),
+    });
+  } catch (err) {
+    handleRestfulError(req, res, next, err);
+  }
+});
+
+router.delete('/nft/mint-subscription/:id', async (req, res, next) => {
+  const { id: subscriptionId } = req.params;
```
williamchong, line 93: Should we require email as input param to verify also?

src/server/api/routes/nft/mint-subscription.js
nwingt: Can remove it

src/server/api/routes/nft/mint-subscription.js
nwingt: Sure

src/server/api/routes/nft/mint-subscription.js
```diff
@@ -0,0 +1,102 @@
...
+    if (!querySnapshot.empty) {
+      res.status(409).send('ALREADY_SUBSCRIBED');
+      return;
+    }
+    const subscriptionId = uuidv4();
+    await nftMintSubscriptionCollection.doc(subscriptionId).create({
+      subscriberEmail,
+      subscribedWallet,
+      ts: firestore.FieldValue.serverTimestamp(),
+    });
```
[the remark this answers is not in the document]
nwingt: Added

functions/firestore/index.js
```diff
@@ -0,0 +1,17 @@
...
+
+module.exports = firestore
+  .document(`${config().db.firestore_nft_mint_subscriptions_root}/{id}`)
+  .onCreate(snapshot => {
+    const subscriptionId = snapshot.id;
+    const { subscriberEmail, subscribedWallet } = snapshot.data();
+    return sendEmail({
+      email: subscriberEmail,
+      subject: 'Writing NFT - NFT Subscription',
+      body: `Hi you have subscribed ${subscribedWallet}, you can cancel the subscription by ${EXTERNAL_URL}/${subscribedWallet}/unsubscribe/${subscriptionId}`,
```
williamchong, line 15: Can we somehow make a global token for a email to unsubscribe all wallet?

functions/modules/ses.js
```diff
@@ -0,0 +1,34 @@
+const aws = require('aws-sdk');
+const functions = require('firebase-functions');
+
+aws.config.accessKeyId = functions.config().aws.config.access_key_id;
+aws.config.secretAccessKey = functions.config().aws.config.secret_access_key;
+aws.config.region = functions.config().aws.config.region;
+
+const ses = new aws.SES();
```
williamchong, line 8: We prefer sendgrid since email is not verified

src/server/api/routes/nft/mint-subscription.js
```diff
@@ -70,10 +78,31 @@ router.get('/nft/mint-subscription', async (req, res, next) => {
...
+        return;
+      }
+
+      const { subscriberEmail } = doc.data();
+      if (email !== subscriberEmail) {
+        res.status(404).send('SUBSCRIPTION_NOT_FOUND');
+        return;
+      }
+
+      await t.delete(docRef);
```
williamchong, line 103: No need for transaction in `delete()` if it works anyway with `try catch` without `get()`

williamchong (changes requested):

functions/modules/ses.js
williamchong: Btw since we are sending a welcome email anyway, do we want to include a `verify` link to prevent spam?
We then will only send email to verified email later.

nwingt: Do you mean adding a new page & API to let users verify?

williamchong: Yes, since we are sending email anyway, put the verification CTA and move the `thank you and you will be notified` to verification page, and let email to be verified for each subscription seems not much of UX issue.
This probably is extra and not required for now though, maybe we want to use the account system later on and just verify once per user

functions/firestore/index.js
nwingt: Maybe we implement this in session feature?

src/server/api/routes/nft/mint-subscription.js
```diff
@@ -0,0 +1,145 @@
...
+      });
+      res.json({ subscriptionId });
+    });
+  } catch (err) {
+    handleRestfulError(req, res, next, err);
+  }
+});
+
+router.get('/nft/mint-subscription', async (req, res, next) => {
+  const { wallet: subscribedWallet, email: subscriberEmail } = req.query;
```
williamchong, line 53: We probably want to remove this endpoint after we have login sessions and user email in place. This is still kind of leak of information if I can just query status of an email + wallet pair

src/server/api/routes/nft/mint-subscription.js
```diff
@@ -0,0 +1,145 @@
...
+      }
+
+      const subscriptionId = uuidv4();
+      const docRef = nftMintSubscriptionCollection.doc(subscriptionId);
+      await t.set(docRef, {
+        subscriberEmail,
+        subscribedWallet,
+        ts: firestore.FieldValue.serverTimestamp(),
+      });
+      res.json({ subscriptionId });
```
williamchong, line 45: res.json should happen outside of `db.runTransaction`

src/server/api/routes/nft/mint-subscription.js
```diff
@@ -0,0 +1,145 @@
...
+        return;
+      }
+
+      const { subscriberEmail } = doc.data();
+      if (email !== subscriberEmail) {
+        res.status(404).send('SUBSCRIPTION_NOT_FOUND');
+        return;
+      }
+
+      await t.set(docRef, { isVerified: true }, { merge: true });
```
williamchong, line 127: using plain `update()` would have the same effect, without using `runTransaction` (update throws on not exists)

src/server/api/routes/nft/mint-subscription.js
```diff
@@ -0,0 +1,145 @@
...
+    await db.runTransaction(async t => {
+      const doc = await t.get(docRef);
+      if (!doc.exists) {
+        res.status(404).send('SUBSCRIPTION_NOT_FOUND');
+        return;
+      }
+
+      const { subscriberEmail } = doc.data();
+      if (email !== subscriberEmail) {
+        res.status(404).send('SUBSCRIPTION_NOT_FOUND');
```
williamchong, line 123: Should throw error with message in transaction, then handle with `res.status`. Avoid calling non-transaction related function in `db.runTransaction`

src/server/api/routes/nft/mint-subscription.js
```diff
@@ -0,0 +1,145 @@
...
+    });
+  } catch (err) {
+    handleRestfulError(req, res, next, err);
+  }
+});
+
+router.delete('/nft/mint-subscription/:id', async (req, res, next) => {
+  const { id: subscriptionId } = req.params;
+  try {
+    await nftMintSubscriptionCollection.doc(subscriptionId).delete();
```
williamchong, line 138: Maybe we can allow removing all subscriptions from a same email with a special querystring, to be used in email for unsubscribe all

src/server/api/routes/nft/mint-subscription.js
williamchong: Where is this api used actually?

williamchong (changes requested):

src/server/api/routes/nft/mint-subscription.js
nwingt: The query string value is a secret for each email?

src/server/api/routes/nft/mint-subscription.js
nwingt: Removed

williamchong (approved):

PR #729, opened 2022-10-22
⚡️ Do not trigger mounted call for invisible nft items in portfolio

williamchong:
on portfolio item mount, 3 API will be called
if we just use `v-show` for created and collected items, collected nft classes info will be fetched first even if we are on created page
This pr use placeholder item to allow calculation of layout without triggering nft item fetch

Further improvement would be to actually store the info we get from `mount` into store to prevent repeated fetching on tab switching


nwingt:
Seems user stats need all items fetch

nwingt:
For lazy fetching, how about using `lazy-component` like `<NFTCampaignItem />`?

https://github.com/likecoin/liker-land/blob/22cfcae180f69daa4503dd025a06bab53802e45c/src/components/NFTCampaign/Item.vue#L2-L37

williamchong:
> Seems user stats need all items fetch

Yes but we can't rely on the mounted, otherwise huge uncontrollable blocking time caused by UI items
Need major refactoring..

nwingt:
Using `v-show` is due to a glitch in `<MagicGrid/>` as @AuroraHuang22 said

williamchong:
> For lazy fetching, how about using `lazy-component` like `<NFTCampaignItem />`?
> 
> https://github.com/likecoin/liker-land/blob/22cfcae180f69daa4503dd025a06bab53802e45c/src/components/NFTCampaign/Item.vue#L2-L37

I think we need better control on fetching data instead of relying of mounting of ui, lazy loading cards would still have the same api spam vs user stat data issue

nwingt:
> Yes but we can't rely on the mounted, otherwise huge uncontrollable blocking time caused by UI items
Need major refactoring..

As I suggested before, we should separate the user stats API?

williamchong:
> > Yes but we can't rely on the mounted, otherwise huge uncontrollable blocking time caused by UI items
> > Need major refactoring..
> 
> As I suggested before, we should separate the user stats API?

Yeah probably the simplest way to go for now

williamchong:
Now https://github.com/likecoin/liker-land/pull/751 is merged, we can review this again

nwingt (approved):
I just tested, seems ok

PR #731, opened 2022-10-24
🚸 Blocking update ownerList after transferring NFT

WeiJun0827:


nwingt (commented):
Should show some loading UI while blocking?

WeiJun0827:
> Should show some loading UI while blocking?

I think currently it just extends the processing time a bit, should we explicitly let user know the state?

nwingt:
How about moving that line to the following lines, perhaps using `Promise.all()` with collected count?
https://github.com/likecoin/liker-land/blob/4d3479b8a0d87c618f7e48ee49b88d84b1cc0ef3/src/mixins/nft.js#L604-L605

WeiJun0827:
> How about moving that line to the following lines, perhaps using `Promise.all()` with collected count?
> 
> https://github.com/likecoin/liker-land/blob/4d3479b8a0d87c618f7e48ee49b88d84b1cc0ef3/src/mixins/nft.js#L604-L605

Agreed and updated.

nwingt (approved):

PR #734, opened 2022-10-26
💩 Add simple wnft pubsub handler

williamchong:
The actual pubsub flow has not been setup yet but this is the expected event


functions/pubsub/index.js
```diff
@@ -0,0 +1,53 @@
...
+            // eslint-disable-next-line no-await-in-loop
+            await sendEmail({
+              email: subscriberEmail,
+              subject: `Writing NFT - New NFT by ${sellerWallet}`,
+              html: getBasicTemplate({
+                title: `New NFT Created by ${sellerWallet}`,
+                subtitle: `${classId} is now live`,
+                content: `Go to https://liker.land/nft/class/${classId}`,
+              }),
+            });
```
nwingt, line 42: Should we wrap this with a `try catch` in case one of the sending requests fails and discard the rest?

functions/pubsub/index.js
```diff
@@ -0,0 +1,57 @@
...
+            const { subscriberEmail } = doc.data();
+            try {
+              // eslint-disable-next-line no-await-in-loop
+              await sendEmail({
+                email: subscriberEmail,
+                subject: `Writing NFT - New NFT by ${sellerWallet}`,
+                html: getBasicTemplate({
+                  title: `New NFT Created by ${sellerWallet}`,
+                  subtitle: `${classId} is now live`,
+                  content: `Go to https://liker.land/nft/class/${classId}`,
```
nwingt, line 41:
```suggestion
                  content: `Go to ${EXTERNAL_URL}/nft/class/${classId}`,
```

nwingt (approved):

nwingt (approved):

PR #737, opened 2022-10-28
⬆️ Update firebase deps to use env as function param

williamchong:
using params allow .env to be effective in deploy time, fixing the undefined collection issue
need some testing if upgrade firebase-admin breaks things


williamchong:
Params not working for firestore path now... close for now

PR #743, opened 2022-11-01
⚡️ Use resized nft image url

williamchong:
Requires https://github.com/likecoin/likecoin-api-public/pull/384

### DIFF
--- a/src/components/NFTWidget/ContentPreview.vue
+++ b/src/components/NFTWidget/ContentPreview.vue
@@ -22,6 +22,8 @@
 </template>
 
 <script>
+import { getLikeCoResizedImageUrl } from '~/util/ui';
+
 export default {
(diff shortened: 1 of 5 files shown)


src/components/NFTWidget/ContentPreview.vue
```diff
@@ -51,6 +51,12 @@ export default {
         backgroundColor: this.imgBgColor,
       };
     },
+    imageSize() {
+      return 450;
+    },
+    resizedImageSrc() {
+      return this.imageSrc.replace(/\?size=\d+/, `?size=${this.imageSize}`);
+    },
```
nwingt, line 59: Should refactor this func to util? seems keep repeating

nwingt (approved):

PR #751, opened 2022-11-06
✨ Load userStats from like.co API

WeiJun0827:


williamchong (approved):
Weird that you would have to refetch user stat going from portfolio to share page, but lets start with this PR for now

nwingt (approved):

nwingt (commented):
Seems fetching user stats in `<UserStatsBase/>` instead of portfolio/dashboard/share page is more convenient?

WeiJun0827:
> Seems fetching user stats in `<UserStatsBase/>` instead of portfolio/dashboard/share page is more convenient?

Was thinking of the pattern that letting the component(`<UserStatsBase>` here) fetch data it uses or allowing the page fetch data and then inject it into the component.

src/mixins/portfolio.js
```diff
@@ -214,6 +216,10 @@ export default {
         this.currentTab = 'collected';
       }
     },
+    async updateUserStats(address) {
+      const { data } = await this.$api.get(getUserNFTStats(address));
+      this.userStats = data;
+    },
```
nwingt, line 222: Why not put inside `<UserStatsBase>`?

nwingt: And add a watcher for `statWallet` to refresh the data

nwingt: For this pattern, the parent does not need to know the details of fetch & data structure, however, the parent can set `ref="userStats"` in `<UserStatsBase />`, then fetch the data using `this.$refs.userStats.fetch()`

nwingt: Maybe `<UserStatsBase/>` can rename to `<UserStatsController />`

WeiJun0827: Indeed make sense.

src/components/UserStats/Controller.vue
```diff
@@ -0,0 +1,69 @@
...
+      return this.userStats?.createdCollectorCount || 0;
+    },
+    isLoadingStats() {
+      return !this.userStats;
+    },
+  },
+  watch: {
+    statWallet() {
+      if (statWallet) {
+        this.getUserStats();
```
nwingt, line 53: ??
```suggestion
        this.updateUserStats();
```

src/components/UserStats/Controller.vue
```diff
@@ -0,0 +1,69 @@
...
+export default {
+  mixins: [portfolio],
+  props: {
+    tag: {
+      type: String,
+      default: 'div',
+    },
+    statWallet: {
+      type: String,
+      default: '',
```
nwingt, line 25:
```suggestion
      required: true,
```

nwingt (changes requested):

nwingt (approved):

PR #755, opened 2022-11-07
🐛 Fix unable to collect during first WC connection

nwingt:
Wait for LIKE balance fetch before collecting NFT with LIKE


src/mixins/nft.js
```diff
@@ -440,6 +440,9 @@ export default {
         1
       );
       try {
+        if (this.walletLIKEBalanceFetch) {
+          await this.walletLIKEBalanceFetch;
```
williamchong, line 444: Maybe `walletLIKEBalanceFetching` or `walletLIKEBalanceFetchPromise` ? very confusing now since `walletFetchLIKEBalance` and `walletLIKEBalanceFetch` both exists

src/store/modules/wallet.js
```diff
@@ -22,6 +23,7 @@ const state = () => ({
   isInited: null,
   methodType: null,
   likeBalance: null,
+  likeBalanceFetchPromise: false,
```
williamchong, line 26: `null` is better than `false` since promise would be an object if exists

src/store/modules/actions/wallet.js
```diff
@@ -92,6 +92,10 @@ export async function initIfNecessary({ dispatch }) {
 
 export async function walletFetchLIKEBalance({ commit, state }) {
   const { address } = state;
-  const balance = await getAccountBalance(address);
-  commit(types.WALLET_LIKE_BALANCE, balance);
+  const balanceFetch = getAccountBalance(address);
+  commit(types.WALLET_SET_LIKE_BALANCE_FETCH_PROMISE, balanceFetch);
+  const balance = await balanceFetch;
+  commit(types.WALLET_SET_LIKE_BALANCE, balance);
+  commit(types.WALLET_SET_LIKE_BALANCE_FETCH_PROMISE, undefined);
```
williamchong, line 99: What if `balanceFetch` throws? should we still set `(types.WALLET_SET_LIKE_BALANCE_FETCH_PROMISE, undefined)`?

williamchong (changes requested):

williamchong (approved):

PR #756, opened 2022-11-07
🐛 Fix issue & improve LIKE balance fetch

nwingt:


src/mixins/nft.js
```diff
@@ -442,6 +442,8 @@ export default {
       try {
         if (this.walletLIKEBalanceFetchPromise) {
```
williamchong, line 443: no need to check here if this.walletFetchLIKEBalance already check?

williamchong (approved):

PR #763, opened 2022-11-09
💬 [Mint Subscription Email] Update content & trim display name

nwingt:


williamchong (approved):

functions/modules/utils.js
```diff
@@ -0,0 +1,9 @@
+exports.shortenString = (s, limit = 15, skippedPartReplacer = '...') => {
+  if (!s || s.length <= limit) return s;
+  const trimmedLength = Math.ceil((limit - skippedPartReplacer.length) / 2);
+  const [head, tail] = [
+    s.substring(0, trimmedLength),
```
williamchong, line 5: problematic if `skippedPartReplacer.length` > `limit.length`, but hard to imagine why would it be used like this

PR #770, opened 2022-11-11
✨[Creator Follow] Allow re-send verification email

nwingt:


src/pages/_id/index.vue
```diff
@@ -230,17 +256,37 @@ export default {
...
     return undefined;
   },
+  created() {
+    if (
+      this.currentTab !== 'created' &&
+      this.$route.hash === this.creatorFollowSectionHash
+    ) {
+      // NOTE: `route.hash` never sent to server side,
+      // Therefore, it is done in here
+      // Ref: https://www.rfc-editor.org/rfc/rfc2396#section-4
```
williamchong, line 266: created is also a universal hook that gets called by server side, if we want client only we should use `mounted`

functions/firestore/index.js
```diff
@@ -22,16 +22,27 @@ module.exports = runWith({ secrets: ['SENDGRID_API_KEY'] })
...
+      return Promise.resolve();
+    }
+
+    const {
+      subscriberEmail,
+      subscribedWallet,
+      isVerified,
+    } = change.after.data();
+    if (isVerified) {
+      return Promise.resolve();
```
williamchong, line 36: if these `return Promise.resolve()` is for type consistency, I think awaiting `sendEmail` below before returning, and switch to return primitive types across this function is simpler

src/locales/en.json
```diff
@@ -331,7 +331,7 @@
   "portfolio_subscription_notify_error_alert": "Error occurred when following {creator}: {error}",
   "portfolio_subscribe_title": "Follow {creator}",
   "portfolio_subscribe_message": "Are you sure to confirm following {creator}'s Writing NFT updates by {email}?",
-  "portfolio_subscribe_message_success": "You have successfully subscribed {creator}. You will receive email when {creator} create any Writing NFT.",
+  "portfolio_subscribe_message_success": "You are successfully following {creator}. You will receive email when {creator} create any Writing NFT.",
```
williamchong, line 334: `You are now following {creator}`

src/locales/zh-Hant.json
```diff
@@ -333,7 +333,7 @@
   "portfolio_subscription_notify_error_alert": "Error occurred when following {creator}: {error}",
   "portfolio_subscribe_title": "Follow {creator}",
   "portfolio_subscribe_message": "Are you sure to confirm following {creator}'s Writing NFT updates by {email}?",
-  "portfolio_subscribe_message_success": "You have successfully subscribed {creator}. You will receive email when {creator} create any Writing NFT.",
+  "portfolio_subscribe_message_success": "You are successfully following {creator}. You will receive email when {creator} create any Writing NFT.",
```
williamchong, line 336: `You are now following {creator}`

src/mixins/portfolio.js
```diff
@@ -58,13 +58,18 @@ export default {
       ],
     };
   },
+  fetch({ redirect, query, route }) {
+    if (!query.tab || !['collected', 'created'].includes(query.tab)) {
+      redirect({
+        ...route,
+        query: { ...query, tab: 'collected' },
+      });
```
williamchong, line 66: Would this causes a 302 on first load? Can we use default values and/or `$router.replace` in client side instead?

src/mixins/portfolio.js
```diff
@@ -210,15 +219,18 @@ export default {
...
         this.isLoading = false;
       }
-      if (!this.sortedCreatedClassIds.length) {
-        this.currentTab = 'collected';
-      }
     },
-    goCollected() {
-      this.currentTab = 'collected';
+    changeTab(tab) {
+      this.$router.push({
```
williamchong, line 224: not sure if `replace` should be used here or keep `push`

src/server/api/routes/nft/mint-subscription.js
```diff
@@ -30,11 +31,22 @@ router.post('/nft/mint-subscription', async (req, res, next) => {
...
     const result = await db.runTransaction(async t => {
       const querySnapshot = await t.get(queryRef);
+      let subscriptionId;
       if (!querySnapshot.empty) {
-        throw new Error('ALREADY_SUBSCRIBED');
+        const [doc] = querySnapshot.docs;
+        const { isVerified, ts } = doc.data();
+        if (
+          isVerified ||
+          Date.now() - ts.toMillis() < VERIFICATION_EMAIL_RESEND_COOLDOWN_IN_MS
```
williamchong, line 40: Send a different error message for cooldown case?

williamchong (changes requested):

src/mixins/portfolio.js
nwingt: Seem using replace keep the current UX, will chang this

src/mixins/portfolio.js
nwingt: Seem not causing

williamchong: it won't if we are coming from a likerland page (client mode), but if we land from other site into likerland via this page, server mode would be used, the redirect might cause a 302 since it's server

nwingt: Updated

WeiJun0827 (approved):

williamchong (approved):

PR #774, opened 2022-11-14
♻️ Refactor user info related computed props

nwingt:


src/mixins/user-info.js
```diff
@@ -0,0 +1,38 @@
+import { mapGetters } from 'vuex';
+
+import { getIdenticonAvatar } from '~/util/api';
+
+export const createUserInfoMixin = ({ propKey = 'User' } = {}) => {
+  const getPropName = (label, prefix = '') => {
```
williamchong, line 6: How about just take a template string as param then call .replace? e.g. getPropName('Is{prop}CivicLiker')

src/mixins/user-info.js
```diff
@@ -0,0 +1,38 @@
...
+        );
+      },
+      [getPropName('CivicLiker', 'is')]() {
+        return !!(
+          this[userInfoPropName]?.isCivicLikerTrial ||
+          this[userInfoPropName]?.isSubscribedCivicLiker
+        );
+      },
+      [getPropName('DisplayName')]() {
+        return this[userInfoPropName]?.displayName || this.wallet;
```
williamchong, line 29: Should we default `ellipsis` if wallet?

nwingt:
Rebase to solve conflict

williamchong (approved):

PR #777, opened 2022-11-15
🥅 Catch not found error for NFT purchase info & metadata

nwingt:


src/mixins/nft.js
```diff
@@ -320,11 +320,25 @@ export default {
       }
     },
     async updateNFTClassMetadata() {
-      await this.fetchNFTMetadata(this.classId);
+      try {
+        await this.fetchNFTMetadata(this.classId);
+      } catch (error) {
+        if (error.response?.status !== 404) {
+          // eslint-disable-next-line no-console
+          console.error(error);
```
williamchong, line 328: please use JSON.stringify since axios error is multiline

src/mixins/nft.js
```diff
@@ -320,11 +320,25 @@ export default {
...
       this.updateDisplayNameList(this.iscnOwner);
     },
     async updateNFTPurchaseInfo() {
-      await this.fetchNFTPurchaseInfo(this.classId);
+      try {
+        await this.fetchNFTPurchaseInfo(this.classId);
+      } catch (error) {
+        if (error.response?.status !== 404) {
+          // eslint-disable-next-line no-console
+          console.error(error);
```
williamchong, line 339: please use JSON.stringify since axios error is multiline

williamchong (changes requested):

src/mixins/nft.js

src/mixins/nft.js

williamchong (approved):

PR #780, opened 2022-11-15
✨ Integrate setting email and following creator

WeiJun0827:
- Implement setting email endpoints
- Verify that the user has set email before following/unfollowing a creator


src/server/api/routes/users/v2/email.js
```diff
@@ -0,0 +1,118 @@
...
+          emailSetTs: FieldValue.serverTimestamp(),
+        });
+      });
+      const confirmLink = `${EXTERNAL_URL}/api/v2/users/${user}/email?token=${token}`;
+      await sendEmail({
+        email,
+        subject: 'Confirm your email in Liker.Land',
+        html: `Please click the link to confirm your email: ${confirmLink}`,
+      });
+      res.sendStatus(200);
```
nwingt, line 62: I think can send status 200 first then send email

src/server/api/routes/users/v2/email.js
```diff
@@ -0,0 +1,118 @@
...
+        const { email: currentEmail, emailSetTs } = userDoc.data();
+        if (email === currentEmail) {
+          throw new Error('SAME_EMAIL_ALREADY_SET');
+        }
+        if (
+          emailSetTs &&
+          Date.now() - emailSetTs.toMillis() <
+            VERIFICATION_EMAIL_RESEND_COOLDOWN_IN_MS
+        ) {
+          throw new Error('EMAIL_SET_IN_COOLDOWN');
```
nwingt, line 47:
```suggestion
          throw new Error('EMAIL_UPDATE_IN_COOLDOWN');
```

src/server/api/routes/users/v2/email.js
```diff
@@ -0,0 +1,118 @@
...
+        res.status(400).send('MISSING_EMAIL');
+        return;
+      }
+      const token = uuidv4();
+      await db.runTransaction(async t => {
+        const userRef = walletUserCollection.doc(user);
+        const userDoc = await t.get(userRef);
+        const { email: currentEmail, emailSetTs } = userDoc.data();
+        if (email === currentEmail) {
+          throw new Error('SAME_EMAIL_ALREADY_SET');
```
nwingt, line 40:
```suggestion
          throw new Error('EMAIL_ALREADY_UPDATED');
```

src/server/api/routes/users/v2/email.js
```diff
@@ -0,0 +1,118 @@
...
+          Date.now() - emailSetTs.toMillis() <
+            VERIFICATION_EMAIL_RESEND_COOLDOWN_IN_MS
+        ) {
+          throw new Error('EMAIL_SET_IN_COOLDOWN');
+        }
+
+        await t.update(userRef, {
+          emailUnconfirmed: email,
+          emailVerifyToken: token,
+          emailSetTs: FieldValue.serverTimestamp(),
```
nwingt, line 53:
```suggestion
          emailLastUpdatedTs: FieldValue.serverTimestamp(),
```

src/server/api/routes/users/v2/email.js
```diff
@@ -0,0 +1,118 @@
...
+          res.status(429).send('EMAIL_SET_IN_COOLDOWN');
+          break;
+        default:
+          handleRestfulError(req, res, next, err);
+      }
+    }
+  }
+);
+
+router.get('/v2/users/:wallet/email', async (req, res) => {
```
nwingt, line 78: This confirm email API should use `PUT`/`UPDATE` method instead of GET

src/server/api/routes/users/v2/follow.js
```diff
@@ -42,12 +43,24 @@ router.post(
         res.status(400).send('INVALID_CREATOR_ADDRESS');
         return;
       }
-      await walletUserCollection.doc(user).update({
-        followers: FieldValue.arrayUnion(creator),
+      await db.runTransaction(async t => {
+        const userRef = walletUserCollection.doc(user);
+        const userDoc = await t.get(userRef);
+        const { email } = userDoc.data();
+        if (!email) throw new Error('EMAIL_NOT_SET_YET');
```
williamchong, line 50: Do we need to distinguish better not set and no verified for frontend to take action?

src/server/api/routes/users/v2/email.js
```diff
@@ -0,0 +1,118 @@
...
+  VERIFICATION_EMAIL_RESEND_COOLDOWN_IN_MS,
+} = require('../../../constant');
+const { EXTERNAL_URL } = require('../../../../config/config');
+
+const router = Router();
+
+router.post(
+  '/v2/users/:wallet/email',
+  authenticateV2Login,
+  checkWalletMatch,
```
williamchong, line 25: How can frontend know email of current login user, and its verification status? in `/self`?

src/server/api/routes/users/v2/email.js
williamchong: cloud function goes haywire if you return res first and do await stuff, better keep this order

src/server/api/routes/users/v2/email.js
```diff
@@ -0,0 +1,118 @@
...
+        throw new Error('INVALID_TOKEN');
+      }
+      await t.update(userRef, {
+        email: emailUnconfirmed,
+        emailUnconfirmed: FieldValue.delete(),
+        emailVerifyToken: FieldValue.delete(),
+      });
+      return emailUnconfirmed;
+    });
+    res.json({ email });
```
williamchong, line 103: no need to return email in result?

williamchong (changes requested):

src/server/api/routes/users/v2/email.js
WeiJun0827: Yes, also consider add followers/followees in `/self`?

src/server/api/routes/users/v2/index.js
```diff
@@ -1,12 +1,14 @@
...
 
 const auth = require('./auth');
+const email = require('./email');
 const follow = require('./follow');
 const nfts = require('./nfts');
 
 const router = Router();
 
 router.use(auth);
+router.use(email);
```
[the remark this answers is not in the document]
WeiJun0827: Do you mean maybe we can add a `userInfo` route to handle data that might be edited on the same page? Like `email`, `displayName` etc.?

src/server/api/routes/users/v2/email.js
WeiJun0827: I see, we should use `PUT`/`UPDATE` method to follow the RESTful design.
But can we let users emit a  `PUT`/`UPDATE` request with a simple click on the verification link in an email?

nwingt: The link should be a liker.land page not API endpoint

WeiJun0827: I see, updated

williamchong: We can't just let user see a plaintext OK without any ui after clicking the link, thus there is always a ui page

src/server/api/routes/users/v2/index.js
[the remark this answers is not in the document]
williamchong: email seems somewhere in between, since it's mostly just a get set of private user data, but there is also actual email verification logics
can be it as individual route if no conclusion

src/server/api/routes/users/v2/email.js
williamchong: we might need followers in public api, so it depends
email should be in self though

src/server/api/routes/users/v2/auth.js
```diff
@@ -20,10 +24,11 @@ router.get('/self', authenticateV2Login, async (req, res, next) => {
     setPrivateCacheHeader(res);
     const { user } = req.session;
     const userDoc = await walletUserCollection.doc(user).get();
-    const { displayName, followers } = userDoc.data();
+    const { displayName, eamil, followers } = userDoc.data();
```
williamchong, line 27:
```suggestion
    const { displayName, email, followers } = userDoc.data();
```

src/server/api/routes/users/v2/auth.js
```diff
@@ -20,10 +24,11 @@ router.get('/self', authenticateV2Login, async (req, res, next) => {
...
     res.json({
       user,
       displayName,
+      eamil,
```
williamchong, line 31:
```suggestion
      email,
```

src/server/api/routes/users/v2/follow.js
```diff
@@ -41,12 +47,33 @@ router.post(
...
+        const userDoc = await t.get(userRef);
+        const { email, emailUnconfirmed } = userDoc.data();
+        if (!email) {
+          if (emailUnconfirmed) {
+            throw new Error('EMAIL_UNCONFIRMED');
+          } else {
+            throw new Error('EMAIL_NOT_SET_YET');
+          }
+        }
+        await t.update(userRef, {
```
williamchong, line 61: note that t.update is not an async call

src/server/api/routes/users/v2/email.js
```diff
@@ -0,0 +1,118 @@
...
+      return emailUnconfirmed;
+    });
+    res.sendStatus(200);
+  } catch (error) {
+    switch (error.message) {
+      case 'USER_NOT_FOUND':
+        res.status(404).send('USER_NOT_FOUND');
+        break;
+      case 'INVALID_ID':
+        res.status(400).send('INVALID_ID');
```
williamchong, line 110:
```suggestion
      case 'INVALID_TOKEN':
        res.status(400).send('INVALID_TOKEN');
```

williamchong (changes requested):

src/server/api/routes/users/v2/follow.js
WeiJun0827: I guess unless `t.update()` is the last line in a `runTransaction()`, we should add an `await` for it?
https://firebase.google.com/docs/firestore/manage-data/transactions#passing_information_out_of_transactions

williamchong (approved):
Need to integrate with UI to know if additional changes is needed

PR #783, opened 2022-11-16
✨ Implement new NFT details page

nwingt:
image

### DIFF
--- a/src/pages/nft/class/_classId/_nftId.vue
+++ b/src/pages/nft/class/_classId/_nftId.vue
@@ -36,7 +36,7 @@
           />
         </header>
 
-        <section class="flex flex-col desktop:grid grid-cols-3 gap-[16px]">
+        <section class="flex flex-col desktop:grid grid-cols-3 gap-x-[16px] gap-y-[32px]">
           <!-- NFT Message List -->
           <div v-if="!nftIsNew" class="col-span-2 flex flex-col items-center gap-[24px]">
             <CardV2 class="flex gap-[24px] w-full">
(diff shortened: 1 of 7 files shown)


src/pages/nft/class/_classId/_nftId.vue
```diff
@@ -104,17 +77,86 @@
...
+            <div
+              v-if="nftCreatorMessage"
+              class="py-[40px] px-[24px] w-full"
+            >
+              <NFTMessage
+                class="w-full"
+                type="purchase"
+                from-type="creator"
+                :from-wallet="iscnOwner"
+                :message="nftCreatorMessage"
```
williamchong, line 140: in API nftCreatorMessage is a template string that has a `.replace` element (currently not used), should we implement it in client also?

nwingt: Implement? Isn't the return memo already parsed?

williamchong: but `nftCreatorMessage` is just `return this.NFTClassMetadata?.message || '';` here without parsing?

nwingt: I thought the parsing is done during mint, am I getting it wrong?

nwingt: Wait, I got it, the parsing is done during collect not mint, so this shows the preview of the message, right?

williamchong: yes, parsing is done during transaction and result is written into memo
so if we really start using template string in message, we might need to implement parsing here too
or we can also ignore this too, the template string is kind of readable

nwingt: We should put the logic into one lib when we develop this further

nwingt:
Merge this into `feature/pnft` branch first until we are ready to launch

PR #791, opened 2022-11-21
✨ Support toggling featured/hidden NFTs

WeiJun0827:


src/components/NFTFeatured/index.vue
```diff
@@ -155,22 +149,35 @@ export default {
...
-      switch (this.currentState) {
-        case CurrentState.FEATURED:
-          this.$emit('state-change', CurrentState.HIDDEN);
+      switch (this.displayState) {
+        case NFT_DISPLAY_STATE.FEATURED:
+          this.removeNFTFeatured({
+            address: this.getAddress,
+            classId: this.classId,
+          });
+          this.addNFTHidden({
```
nwingt, line 158: I think the API call below can be moved to the action

src/store/modules/nft.js
```diff
@@ -307,31 +305,31 @@ const actions = {
     classIdSet.add(classId);
     commit(TYPES.NFT_SET_USER_NFT_CLASS_FEATURED_SET_MAP, {
       address,
-      classIdSet,
+      classIdSet: new Set(classIdSet), // clone to trigger reactivity
```
nwingt, line 349: How about using `Vue.set` in the `NFT_SET_USER_NFT_CLASS_FEATURED_SET_MAP` mutation?

src/store/modules/nft.js
```diff
@@ -131,16 +157,31 @@ const getters = {
...
     orderBy,
     order = ORDER.DESC,
+    enableFeaturedAndHidden,
   }) => {
-    const sorted = [...nfts].sort((nA, nB) => {
+    const filtered = enableFeaturedAndHidden
+      ? nfts.filter(
+          nft => !state.userNFTClassHiddenSetMap[nftOwner]?.has(nft.classId)
+        )
+      : [...nfts];
```
nwingt, line 171: I think can refactor the filter logic part into getters
```suggestion
      ? getters.filterNFTClassListWithState(nfts)
      : []
```

src/mixins/nft.js
```diff
@@ -252,14 +252,12 @@ export default {
       return wallet => (wallet === this.iscnOwner ? 'creator' : 'collector');
     },
     nftDisplayState() {
-      if (
-        this.getNFTClassFeaturedSetByAddress(this.getAddress)?.has(this.classId)
-      ) {
+      const address =
+        this.$route.name === 'id' ? this.$route.params.id : this.getAddress;
+      if (this.getNFTClassFeaturedSetByAddress(address)?.has(this.classId)) {
```
nwingt: `this.wallet` is the address
```suggestion
      if (this.getNFTClassFeaturedSetByAddress(this.wallet)?.has(this.classId)) {
```

src/mixins/nft.js
```diff
@@ -252,12 +252,17 @@ export default {
...
     },
     nftDisplayState() {
-      const address =
+      // should use the address in URL as the subject address when browsing other's profile
+      const subjectAddress =
         this.$route.name === 'id' ? this.$route.params.id : this.getAddress;
-      if (this.getNFTClassFeaturedSetByAddress(address)?.has(this.classId)) {
+      if (
+        this.getNFTClassFeaturedSetByAddress(subjectAddress)?.has(this.classId)
+      ) {
```
nwingt, line 260: Same as above

src/store/modules/nft.js
WeiJun0827: Should it be a function like this?
```
filterNFTClassListWithState: state => nfts =>
    nfts.filter(
      nft => !state.userNFTClassHiddenSetMap[collector]?.has(nft.classId)
    ),
```

src/store/modules/nft.js
WeiJun0827: Already used `Vue.set()` in [mutation](https://github.com/likecoin/liker-land/pull/791/files#diff-bc31414c5136bf8f64828fe6722217f2a2dc6a0e66fc238cce705547403f6dc2R38-R49). Seems VueX only triggers UI reactivity when the pointer in the field of object state (`state.userNFTClassFeaturedSetMap`) changed.

WeiJun0827:
Migrate this PR to https://github.com/likecoin/liker-land/pull/793

src/store/modules/nft.js
nwingt: Yes seem so

src/mixins/nft.js
WeiJun0827: Which mixin should we put a factory function in to let two pages import? Portfolio mixin?

PR #793, opened 2022-11-22
✨ Support toggling featured/hidden NFTs

WeiJun0827:
Migrate https://github.com/likecoin/liker-land/pull/791 to this PR


src/pages/nft/class/_classId/_nftId.vue
```diff
@@ -114,6 +114,11 @@
                 @collect="handleCollectFromPreviewSection"
                 @view-content="handleViewContent"
               />
+              <NFTFeatured
+                :class-id="classId"
+                :preset="nftCollectorWalletAddress && nftCollectorWalletAddress.includes(getAddress) ? 'edit' : 'invisible'"
```
williamchong, line 119: confusing preset naming, can we v-if this whole component instead inside if it needs to be hidden?

src/components/NFTFeatured/index.vue
```diff
@@ -0,0 +1,176 @@
...
+    ...mapActions([
+      'signLogin',
+      'addNFTFeatured',
+      'addNFTHidden',
+      'removeNFTFeatured',
+      'removeNFTHidden',
+    ]),
+    async handleClick(event) {
+      event.preventDefault();
+      event.stopPropagation();
```
williamchong, line 138: what would happen if we don't stop propagation?

src/components/NFTFeatured/index.vue
```diff
@@ -0,0 +1,176 @@
...
+        'w-[20px]',
+        'absolute',
+        'top-[50%]',
+        'left-[50%]',
+        'translate-x-[-50%]',
+        'translate-y-[-50%]',
+      ];
+    },
+    hasLogin() {
+      return !!this.$cookie.get(AUTH_COOKIE_NAME);
```
williamchong, line 124: Does this even work? if it works it seems to be a security issue

williamchong (changes requested):

src/components/NFTFeatured/index.vue
WeiJun0827: Deprecated code, removed

src/pages/nft/class/_classId/_nftId.vue
WeiJun0827: Modified with the `readOnly` prop, since some CSS classes and props should reference it.

src/store/modules/wallet.js
```diff
@@ -168,6 +182,46 @@ const actions = {
...
+        signature: { signature, pub_key: publicKey },
+      } = await state.signer.sign(state.address, payload);
+      const data = {
+        signature,
+        publicKey: publicKey.value,
+        message: stringify(message),
+        from: state.address,
+      };
+      await this.$api.post(postUserV2Login, data);
+      commit(types.WALLET_SET_HAS_LOGGED_IN, true);
```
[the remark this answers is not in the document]
WeiJun0827: updated

WeiJun0827:
截圖 2022-11-23 下午1 28 08

Got a strange 500 error `Cannot convert undefined or null to object` in [this line](https://github.com/likecoin/liker-land/blob/feature/featured/src/store/modules/wallet.js#L108-L110). Need further investigate.

WeiJun0827:
截圖 2022-11-23 下午1 32 49

Got a strange 500 error Cannot convert undefined or null to object in [this line](https://github.com/likecoin/liker-land/blob/feature/featured/src/store/modules/wallet.js#L108-L110) when switching between Keplr wallets. Need further investigation.

src/components/NFTFeatured/index.vue
nwingt: I think the reason is that clicking `<NFTPortfolioItem/>` will go to the NFT details page, and the `<NFTFeatured/>` is inside, `<NFTPortfolioItem/>`, so `stopPropagation` will prevent going to the NFT details page when clicking `<NFTFeatured/>`.

nwingt:
> Got a strange 500 error Cannot convert undefined or null to object in [this line](https://github.com/likecoin/liker-land/blob/feature/featured/src/store/modules/wallet.js#L108-L110) when switching between Keplr wallets. Need further investigation.

@WeiJun0827 Should be fixed in [9649d3d](https://github.com/likecoin/liker-land/pull/793/commits/9649d3d5cdc47d2bc19e47b39cde5332c5c62f32)

src/components/NFTFeatured/index.vue
```diff
@@ -129,7 +129,7 @@ export default {
       event.stopPropagation();
       if (this.readOnly) return;
 
-      if (!this.hasLoggedIn) {
+      if (!this.loginAddress || this.loginAddress !== this.getAddress) {
```
williamchong, line 132: This can be a getter in store instead?

williamchong (changes requested):

src/components/NFTFeatured/index.vue
nwingt: Possible getter name: `walletHasLoggedIn`

williamchong (approved):

PR #797, opened 2022-11-23
🐛 Filter failed Tx in NFT history

WeiJun0827:


williamchong (approved):
Is is possible not to set failed tx in db? Where was it set in db?

nwingt (approved):

PR #798, opened 2022-11-23
🥅 Detect failed Tx

WeiJun0827:
### Related issues
- https://github.com/likecoin/liker-land/pull/797 
- https://github.com/likecoin/likecoin-api-public/pull/399

### DIFF
--- a/src/mixins/nft.js
+++ b/src/mixins/nft.js
@@ -514,7 +514,7 @@ export default {
           this.classId,
           1
         );
-        const txHash = await broadcastTx(signData, this.getSigner);
+        const { txHash, code } = await broadcastTx(signData, this.getSigner);
         logTrackerEvent(
           this,
           'NFT',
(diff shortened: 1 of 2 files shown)


src/mixins/nft.js
```diff
@@ -514,14 +514,15 @@ export default {
...
-        const txHash = await broadcastTx(signData, this.getSigner);
+        const { txHash, code } = await broadcastTx(signData, this.getSigner);
         logTrackerEvent(
           this,
           'NFT',
           'NFTCollectBroadcastTxComplete',
           this.classId,
           1
         );
+        if (code) throw new Error(`TX_FAILED_WITH_CODE_${code}`);
```
nwingt, line 525: I think we should explicitly check for `0`
```suggestion
        if (code !== 0) throw new Error(`TX_FAILED_WITH_CODE_${code}`);
```

nwingt (approved):

PR #801, opened 2022-11-23
🐛 Fix duplicated key issue

WeiJun0827:
Since we may have `mint_nft` and `new_class` events in the same Tx.
The key should be multiple fields.

### DIFF
--- a/src/mixins/nft.js
+++ b/src/mixins/nft.js
@@ -371,10 +371,10 @@ export default {
           const historyInDB = data.list;
           const eventMap = new Map();
           historyOnChain.forEach(e => {
-            eventMap.set(`${e.txHash}-${e.nftId}`, e);
+            eventMap.set(`${e.txHash}-${e.nftId}-${e.event}`, e);
           });
           historyInDB.forEach(e => {
-            const key = `${e.txHash}-${e.nftId}`;
+            const key = `${e.txHash}-${e.nftId}-${e.event}`;
             if (eventMap.has(key)) {
               eventMap.get(key).price = e.price;
             }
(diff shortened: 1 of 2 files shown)


src/mixins/nft.js
```diff
@@ -371,10 +371,10 @@ export default {
           const historyInDB = data.list;
           const eventMap = new Map();
           historyOnChain.forEach(e => {
-            eventMap.set(`${e.txHash}-${e.nftId}`, e);
+            eventMap.set(`${e.txHash}-${e.nftId}-${e.event}`, e);
```
williamchong, line 374: Do we really want to key each `event`, or do we just want `message` keys?
If its just `message` maybe `msg_index` will do

nwingt (approved):

williamchong (commented):
In what case do we have `mint_nft` and `new_class` in a same transaction? It sounds impossible since we don't have a `class_id` for `mint_nft` if `new_class` is not passed as a transaction first..

src/mixins/nft.js
nwingt: So, the key would be `{txHash}-{msgIndex}`?

WeiJun0827: > In what case do we have `mint_nft` and `new_class` in a same transaction? It sounds impossible since we don't have a `class_id` for `mint_nft` if `new_class` is not passed as a transaction first..

Sorry for typo, should be `mint_nft` and `/cosmos.nft.v1beta1.MsgSend`, not `new_class`.

williamchong: Oh I looked again and the event api only returns event name not msg_index
So current `event` key is good

PR #806, opened 2022-11-25
✨ Fetch nfts metadata in nft details page

williamchong:
Fetch and use nfts metadata if exists in detail page
Not stored in store for now

### DIFF
--- a/src/mixins/nft.js
+++ b/src/mixins/nft.js
@@ -292,7 +292,7 @@ export default {
     ...mapActions([
       'lazyGetUserInfoByAddress',
       'fetchNFTPurchaseInfo',
-      'fetchNFTMetadata',
+      'fetchNFTClassMetadata',
       'fetchNFTOwners',
       'initIfNecessary',
       'uiToggleCollectModal',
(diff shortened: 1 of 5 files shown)


nwingt (commented):
NFTs seems related to multiple NFTs, not individual
In this case, I prefer replacing NFTs to NFT

If NFT is the first word, please me use lower case for naming

williamchong:
> NFTs seems related to multiple NFTs, not individual In this case, I prefer replacing NFTs to NFT
> 
> If NFT is the first word, please me use lower case for naming

used NFTs because for some reason the api is called nfts in lcd...
Any good name for `NFTsExternalUrl`...etc thats shadows `NFTExternalUrl` ...etc if `NFTs` is not to be used?

nwingt:
> > NFTs seems related to multiple NFTs, not individual In this case, I prefer replacing NFTs to NFT
> > If NFT is the first word, please me use lower case for naming
> 
> used NFTs because for some reason the api is called nfts in lcd... Any good name for `NFTsExternalUrl`...etc thats shadows `NFTExternalUrl` ...etc if `NFTs` is not to be used?

I see that API is called `classes` in LCD for NFT Class, I want to keep it consistent

I think `nftExternalURL` is acceptable for now, seems most of the existing fields are `NFTXxxxxx` pattern
Should rename `NFTXXXX` to `nftClassXXXX` later

nnkken (approved):

nwingt:
As this PR should release along with the `develop` branch with updates for NFT Book, I will cherry-pick and open another PR to `develop` and solve the conflicts

PR #807, opened 2022-11-25
✨ [NFT Details] Override NFT Class metadata with NFT metadata

nwingt:
Continue from #806


src/pages/nft/class/_classId/_nftId.vue
```diff
@@ -318,31 +338,33 @@ export default {
...
-        store.dispatch('fetchNFTMetadata', classId),
+        store.dispatch('fetchNFTClassMetadata', classId),
         store.dispatch('lazyGetNFTPurchaseInfo', classId).catch(err => {
           if (err.response?.data !== 'NFT_CLASS_NOT_FOUND') {
             // eslint-disable-next-line no-console
             console.error(JSON.stringify(err));
           }
         }),
       ]);
+      nftMetadata = await store.dispatch('fetchNFTMetadata', {
```
williamchong, line 364: Not sure if we should/can put this into the above `Promise.all`

nwingt: The original action depends on NFT Class metadata for fallback, I think we can resolve that with computed properties.

src/store/modules/nft.js
```diff
@@ -28,6 +29,9 @@ const mutations = {
   [TYPES.NFT_SET_NFT_CLASS_METADATA](state, { classId, metadata }) {
     Vue.set(state.metadataByClassIdMap, classId, metadata);
   },
+  [TYPES.NFT_SET_NFT_METADATA](state, { nftId, metadata }) {
+    Vue.set(state.metadataByNFTIdMap, nftId, metadata);
```
williamchong, line 33: class id +  nft id is the unique key, so we need to set class id too

src/store/modules/nft.js
```diff
@@ -84,6 +88,7 @@ const getters = {
       (acc, val) => acc + val.length,
       0
     ),
+  getNFTMetadataById: state => id => state.metadataByNFTIdMap[id],
```
williamchong, line 91: class id + nft id is the unique key, so we need to pass class id too

williamchong (changes requested):

src/store/modules/nft.js
nwingt: Updated

PR #808, opened 2022-11-26
🩹 Fix event list table duplicated keys

nwingt:


williamchong (approved):
do we want class id if we are adding nftid anyway

nwingt:
But the events of this page is under the same NFT Class, so it makes no different?

williamchong:
> But the events of this page is under the same NFT Class, so it makes no different?

Not for now, depends if you think this component would be used in some cross class id scenerio

williamchong (approved):

PR #811, opened 2022-11-29
💄 [Transfer] Show displayName instead of address

AuroraHuang22:
截圖 2022-11-30 下午1 08 40

截圖 2022-11-30 下午1 10 05

### DIFF
--- a/src/components/EventModal/Transfer.vue
+++ b/src/components/EventModal/Transfer.vue
@@ -103,7 +103,7 @@
       <FormField
         class="mx-[8px]"
         :label="$t('tx_modal_label_receiver')"
-      >{{ toWallet }}</FormField>
+      >{{ toDisplayName }}</FormField>
     </div>
     <template
       v-if="!isTransferring"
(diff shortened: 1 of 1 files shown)


src/components/EventModal/Transfer.vue
```diff
@@ -189,13 +191,16 @@ export default {
...
     handleInputToWallet(value) {
       this.toWallet = value;
     },
     handleInputMemo(value) {
       this.memo = value;
     },
-    handleClickTransfer() {
+    async handleClickTransfer() {
+      const userInfo = await this.fetchUserInfoByAddress(this.toWallet);
+      this.recipientName = userInfo.displayName;
```
nwingt, line 203: - Please use the `lazyGetUserInfoByAddress` action for better caching
- I think it is better to use a computed property e.g. `toDisplayName` and get the display name with the getter `getUserInfoByAddress(toWallet)`

nwingt (changes requested):

nwingt (approved):

PR #817, opened 2022-12-02
💄 Use gem color for card hover effect

williamchong:


src/components/NFTGem/Wrapper.vue
```diff
@@ -44,6 +44,9 @@ export default {
       }
     },
   },
+  created() {
+    this.$emit('set-gem-level', this.gemLevel);
```
nwingt, line 48: Can provide slot props for children component to access gem level

Ref:
https://github.com/likecoin/liker-land/blob/a9b5fc4ec8e9f5fcf59296027c88e9aa58f0c767/src/components/UserStats/Controller.vue#L1-L11

https://github.com/likecoin/liker-land/blob/a9b5fc4ec8e9f5fcf59296027c88e9aa58f0c767/src/components/UserStats/MyDashboard.vue#L2-L27

src/components/NFTPortfolio/Card.vue
```diff
@@ -19,3 +19,24 @@
...
+      type: Number,
+      default: -1,
+    },
+  },
+  computed: {
+    hoverClass() {
+      if (this.gemLevel < 0 || this.gemLevel > NFT_GEM_NAME.length) {
+        return 'border-like-cyan-light';
+      }
+      return `border-gem-${NFT_GEM_NAME[this.gemLevel].toLowerCase()}`;
```
nwingt, line 38: I think it is more simple to just return `border-[#xxxxxx]` for each gem level without changing `tailwind.config.js`

williamchong: tailwind won't define the css for these classes if I dynamically return it in function though, since it can't see them in source code. Or is there any trick to make them work even in dynamic cases?

nwingt: Don't construct class names dynamically
Ref: https://tailwindcss.com/docs/content-configuration#dynamic-class-names

I think just returning the complete class name is more simple
```js
const gemKey = NFT_GEM_NAME[this.gemLevel].toLowerCase();
switch (gemKey) {
  case 'dust':
    return 'hover:border-[#f7f7f7]';

  case 'stone':
  case 'rock':
  case 'boulder':
    return 'hover:border-[#ebebeb]';
  ...
}
```

nwingt: I think it is more elegant to place `hoverClass` prop inside `<NFTGemWrapper>` and return it through slot props and set it to the class of `<NFTPortfolioCard/>` inside `<NFTPortfolioBase/>`

src/components/NFTGem/Wrapper.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div class="relative w-full">
-    <slot />
+    <slot :gemLevel="gemLevel" />
```
nwingt, line 3: Prefer using kebab case for props, ref: https://vuejs.org/guide/components/props.html#prop-name-casing
```suggestion
    <slot :gem-level="gemLevel" />
```

nwingt (changes requested):

nwingt (approved):

PR #821, opened 2022-12-03
💄 [Portfolio] Mitigate items flashing when loading

nwingt:
### Before
https://user-images.githubusercontent.com/17264716/205434571-31cb2c3b-f04c-4cc1-8641-3e7c2a0b156d.mov

### After
https://user-images.githubusercontent.com/17264716/205434576-9b3894cc-6f73-4fef-a62c-6a83ec9246d6.mov

### DIFF
--- a/src/util/nft.js
+++ b/src/util/nft.js
@@ -9,6 +9,7 @@ import {
   LIKECOIN_CHAIN_NFT_RPC,
   LIKECOIN_CHAIN_MIN_DENOM,
   LIKECOIN_NFT_API_WALLET,
+  LIKECOIN_NFT_HIDDEN_ITEMS,
 } from '../constant';
 
 let queryClient = null;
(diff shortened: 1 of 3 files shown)


src/util/nft.js
```diff
@@ -175,6 +176,12 @@ export function checkIsWritingNFT(classMetadata) {
   );
 }
 
+export function checkIsWritingNFTByNFTId(id) {
+  return /^writing-[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i.test(
```
williamchong, line 180: should add comment stating this is kind of performance hack, since wnft should be judged from class metadata

williamchong (approved):

src/util/nft.js
nwingt: Added

PR #824, opened 2022-12-05
💄 Use gemColorClasses for hover color

williamchong:
using same color for bottom and hover border makes the effect much smoother


nwingt:
I rebased the master & develop branch so that gem color hover is not leak to production

PR #836, opened 2022-12-10
🚸 Improvement for wallet connection

nwingt:


src/mixins/nft.js
```diff
@@ -546,6 +547,10 @@ export default {
       } catch (error) {
         this.uiSetTxError(error.response?.data || error.toString());
         this.uiSetTxStatus(TX_STATUS.FAILED);
+      } finally {
+        this.$router.push({
+          query: { ...this.$route.qeury, action: undefined },
```
williamchong, line 552: qeury should be query

williamchong (changes requested):

src/mixins/nft.js
nwingt: Fixed

williamchong (approved):

PR #839, opened 2022-12-14
♻️ Combine featured/hidden APIs and log events

WeiJun0827:


src/server/api/routes/users/v2/nfts/featured.js
```diff
@@ -44,27 +63,67 @@ router.post(
       await walletUserCollection.doc(user).update({
         featuredNFTClassIds: FieldValue.arrayUnion(classId),
```
nwingt, line 64: should remove hidden as well?

src/server/api/routes/users/v2/nfts/featured.js
```diff
@@ -44,27 +63,67 @@ router.post(
...
+    try {
+      setPrivateCacheHeader(res);
+      const { wallet: user } = req.params;
+      const { classId } = req.body;
+      if (!classId) {
+        res.status(400).send('CLASS_ID_MISSING');
+        return;
+      }
+      await walletUserCollection.doc(user).update({
+        hiddenNFTClassIds: FieldValue.arrayRemove(classId),
```
nwingt, line 121: should remove featured as well?

src/server/api/routes/users/v2/nfts/featured.js
```diff
@@ -44,27 +63,67 @@ router.post(
...
       if (!classId) {
         res.status(400).send('CLASS_ID_MISSING');
         return;
       }
       await walletUserCollection.doc(user).update({
         featuredNFTClassIds: FieldValue.arrayRemove(classId),
+        hiddenNFTClassIds: FieldValue.arrayUnion(classId),
+      });
+      publisher.publish(PUBSUB_TOPIC_MISC, req, {
+        logType: 'UserNFTClassHidden',
```
nwingt, line 96:
```suggestion
        logType: 'UserNFTClassHide',
```

src/server/api/routes/users/v2/nfts/featured.js
```diff
@@ -44,27 +63,67 @@ router.post(
...
+      const { classId } = req.body;
+      if (!classId) {
+        res.status(400).send('CLASS_ID_MISSING');
+        return;
+      }
+      await walletUserCollection.doc(user).update({
+        hiddenNFTClassIds: FieldValue.arrayRemove(classId),
+      });
+      publisher.publish(PUBSUB_TOPIC_MISC, req, {
+        logType: 'UserNFTClassUnhidden',
```
nwingt, line 124:
```suggestion
        logType: 'UserNFTClassUnhide',
```

src/server/api/routes/users/v2/nfts/featured.js
```diff
@@ -44,27 +63,67 @@ router.post(
       await walletUserCollection.doc(user).update({
         featuredNFTClassIds: FieldValue.arrayUnion(classId),
       });
+      publisher.publish(PUBSUB_TOPIC_MISC, req, {
+        logType: 'UserNFTClassFeatured',
```
nwingt, line 67:
```suggestion
        logType: 'UserNFTClassFeature',
```

src/server/api/routes/users/v2/nfts/featured.js
```diff
@@ -44,27 +63,67 @@ router.post(
...
+      });
+      res.sendStatus(200);
+    } catch (err) {
+      handleRestfulError(req, res, next, err);
+    }
+  }
+);
+
+router.post(
+  '/:wallet/nfts/unhidden',
```
nwingt, line 108: I think the above 3 API can be handled with a single path `'/:wallet/nfts/:state'`

There are 3 states, `featured`/`hidden`/`normal`

src/components/NFTFeatured/index.vue
```diff
@@ -121,10 +121,9 @@ export default {
   methods: {
     ...mapActions([
       'signLogin',
-      'addNFTFeatured',
-      'addNFTHidden',
-      'removeNFTFeatured',
-      'removeNFTHidden',
+      'setNFTFeatured',
+      'setNFTHidden',
+      'setNFTUnhidden',
```
nwingt, line 126: Maybe combine these actions into one, so you can call and handle the logic the same way in backend
```js
this.setNFTDisplayState({
  state: 'hidden', // 'featured', 'normal'
  address: this.getAddress,
  classId: this.classId,
})
```

src/util/api/index.js
```diff
@@ -269,3 +269,6 @@ export const formatFeaturedNFTUrl = wallet =>
 
 export const formatHiddenNFTUrl = wallet =>
   `/api/v2/users/${wallet}/nfts/hidden`;
+
+export const formatUnhiddenNFTUrl = wallet =>
+  `/api/v2/users/${wallet}/nfts/normal`;
```
nwingt, line 274: I suggest combining the above three functions
```suggestion
export const getNFTDisplayStateURL = (wallet, state) =>
  `/api/v2/users/${wallet}/nfts/${state}`;
```

src/components/NFTFeatured/index.vue
nwingt: You may probably want to combine the mutations`NFT_SET_USER_NFT_CLASS_FEATURED_SET_MAP`
`NFT_SET_USER_NFT_CLASS_HIDDEN_SET_MAP` as well

nwingt (approved):

src/store/modules/nft.js
```diff
@@ -47,17 +46,15 @@ const mutations = {
...
-  [TYPES.NFT_SET_USER_NFT_CLASS_HIDDEN_SET_MAP](
-    state,
-    { address, classIdSet }
-  ) {
-    Vue.set(state.userNFTClassHiddenSetMap, address, classIdSet);
+    Vue.set(
+      state.userNFTClassDisplayStateSetsMap,
+      address,
+      nftDisplayStateSets
+    );
```
williamchong, line 57: We should explicitly define featuredClassIdSet and hiddenClassIdSet here, since we dont just want any arbitrary member and  they are expected everywhere

src/components/NFTFeatured/index.vue
```diff
@@ -144,16 +138,11 @@ export default {
...
-            }),
-            this.addNFTHidden({
-              address: this.getAddress,
-              classId: this.classId,
-            }),
-          ]);
+          await this.setNFTDisplayState({
+            address: this.getAddress,
+            classId: this.classId,
+            displayState: NFT_DISPLAY_STATE.HIDDEN,
```
williamchong, line 144: We dont have a unfeature action, just hide and unhide and feature?

williamchong (changes requested):
Is unhide also unfeature too? if yes should we just call it default or reset?

nwingt:
The state cycles as: featured -> hidden -> normal -> featured...and so on

williamchong:
> The state cycles as: featured -> hidden -> normal -> featured...and so on

unhide as an action should be called reset instead of unhide i guess

src/store/modules/nft.js
williamchong: I mean we should not directly store nftDisplayStateSets here, we should explicitly destruct it into the two member we need, and explicitly store the 2 member, such that it is clear that we want these two member in this mutations

williamchong: We can, or if you want to limit the diff in this PR to only hide/feature related code, maybe in later PRs

williamchong (approved):

PR #844, opened 2022-12-14
🩹 [Portfolio] Hide empty text in creator follow form

nwingt:
As the form now appears on all tabs, it is weird to show hint for no created NFT if the tab has nothing

E.g. Have  collected NFT but no created NFT
### Before
image
### After
image

### DIFF
--- a/src/pages/_id/index.vue
+++ b/src/pages/_id/index.vue
@@ -59,7 +59,7 @@
           class="w-full"
           :creator-wallet-address="wallet"
           :creator-display-name="userDisplayName"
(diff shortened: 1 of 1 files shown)


williamchong (commented):
So is-empty is never used now? Should it be `!nftClassListOfCreatedInOrder.length && connectedWallet === id` or in portfolio page?

williamchong:
> @williamchong Maybe simply remove `is-empty` props for now? I don't know where to put this copywriting image

i think we can show this message and hide the email box if its the creator own self?
but we can merge this pr first

PR #845, opened 2022-12-15
⚡️ Prevent layout shifting after loading in civic dashboard

williamchong:
Before:
image

After:
image

### DIFF
--- a/src/components/CivicLikerV3/Dashboard.vue
+++ b/src/components/CivicLikerV3/Dashboard.vue
@@ -1,27 +1,29 @@
 <template>
-  <Transition name="fade" mode="out-in">
-    <div
-      v-if="!validatorAddress"
-      key="loading"
-      class="flex items-center justify-center min-h-[180px]"
-    >
-      <Spinner class="mx-auto" />
-    </div>
-    <CivicLikerV3PureDashboard
-      v-else
-      key="dashboard"
-      :status="status"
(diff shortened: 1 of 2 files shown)


src/components/CivicLikerV3/Dashboard.vue
```diff
@@ -3,7 +3,7 @@
     <div
       v-if="!validatorAddress"
       key="loading"
-      class="flex items-center justify-center min-h-[180px]"
+      class="flex items-center justify-center min-h-[720px]"
```
nwingt, line 6: 180px is for the height of the banner
image

So when the loading is completed, the spinner will looks like turning into the avatar
https://user-images.githubusercontent.com/17264716/207918815-1fa8a441-1b4c-4a38-b81a-6f5db975eca1.mov

I suggest wrapping the `<Transition/>` with `<div class="min-h-[720px]">` instead

nwingt: Also, I suggest providing `height:64px` for the Civic Liker logo here to prevent slightly shifting 
https://github.com/likecoin/liker-land/blob/a6749be921c56175359768f275b4641d8df02e20/src/components/CivicLikerV3/Page.vue#L9

williamchong: updated, please check if I understood correctly

nwingt (approved):

PR #847, opened 2022-12-19
🔍️. Add canonical link and likecoin meta

williamchong:


src/pages/_id/index.vue
```diff
@@ -144,6 +149,7 @@ export default {
           body: true,
         },
       ],
+      link: [{ rel: 'canonical', href: `/${this.wallet}` }],
```
nwingt, line 152: Need absolute path?
Ref: https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls?visit_id=638070494602143764-1693442656&hl=en&rd=2#rel-canonical-link-method

williamchong: seems its working for existing path though, anyway will add host

PR #848, opened 2022-12-21
🐛 Fix portfolio issues

nwingt:


src/mixins/nft.js
```diff
@@ -324,7 +324,6 @@ export default {
     getAddress(newAddress) {
       if (newAddress) {
         this.fetchUserCollectedCount();
-        this.fetchUserNFTDisplayStateList();
```
nwingt, line 327: As we already fetched in here when changing portfolio/dashboard address
https://github.com/likecoin/liker-land/blob/6626c081eb9b74abb00f747b048773fc1d8dbfd0/src/mixins/portfolio.js#L322-L326

williamchong (approved):

PR #860, opened 2022-12-28
✨ Fetch nft class listing on chain in buy NFT functions

williamchong:


nwingt (approved):

src/store/modules/nft.js
```diff
@@ -272,11 +278,38 @@ const actions = {
...
+        } = l;
+        return {
+          classId,
+          nftId,
+          seller,
+          price: new BigNumber(price).shiftedBy(-9).toNumber(),
+          expiration: new Date(expiration),
+        };
+      })
+      .sort((a, b) => a - b)[0];
```
WeiJun0827, line 302:
```suggestion
      .sort((a, b) => a.price - b.price)[0];
```

src/store/modules/nft.js
```diff
@@ -272,11 +278,38 @@ const actions = {
...
+  },
   async lazyGetNFTPurchaseInfo({ getters, dispatch }, classId) {
     let info = getters.getNFTClassPurchaseInfoById(classId);
+    const listingInfo = getters.getNFTClassListingInfoById(classId);
     if (!info) {
       info = await dispatch('fetchNFTPurchaseInfo', classId);
     }
+    if (!listingInfo) dispatch('fetchNFTListingInfo', classId);
     return info;
   },
```
WeiJun0827, line 314: Seems the `PurchaseInfo` term implies both the original `PurchaseInfo` and the new `ListingInfo`. Would a new term be considered to differentiate the purchase price from the new price that considers the listing price?

src/store/modules/nft.js
```diff
@@ -441,6 +474,7 @@ const actions = {
             catchAxiosError(dispatch('fetchNFTPurchaseInfo', classId))
           );
         }
+        promises.push(dispatch('fetchNFTPurchaseInfo', classId));
```
WeiJun0827, line 477: Can this part be simplified?

src/store/modules/nft.js
williamchong: `lazyGetNFTPurchaseAndListingInfo`?

src/store/modules/nft.js
williamchong: You mean put this to line 465?

WeiJun0827: Yes, will the logic make sense?

src/store/modules/nft.js
WeiJun0827: Yeah, looks good to me

src/store/modules/nft.js
williamchong: Actually this should be `fetchNFTListingInfo`

WeiJun0827 (approved):

PR #863, opened 2023-01-02
✨ [Detail] Display purchase price in history

WeiJun0827:


src/util/nft.js
```diff
@@ -283,6 +285,21 @@ export const getNFTClassesRespectDualPrefix = async (axios, owner) => {
   return arraysOfNFTClasses.flat();
 };
 
+export const getPurchasePrice = async (axios, nftId, txHash) => {
```
williamchong, line 288: should include classId as param since class+nft id is the unique key

williamchong, line 288: should include classId as param since class+nft id is the unique key

src/mixins/nft.js
```diff
@@ -441,7 +442,22 @@ export default {
...
       });
-      let history = historyOnChain;
+
+      const eventMap = new Map();
+      historyOnChain.forEach(e => {
+        eventMap.set(`${e.txHash}-${e.nftId}-${e.event}`, e);
+      });
+
+      const purchaseEvents = historyOnChain.filter(e => e.event === 'buy_nft');
+      const purchasePrices = await Promise.all(
```
williamchong, line 452: can we await this together with getHFTHistory to reduce blocking time?

williamchong (changes requested):

src/util/nft.js
```diff
@@ -283,6 +285,68 @@ export const getNFTClassesRespectDualPrefix = async (axios, owner) => {
...
+      })
+    );
+
+    const historyInDB = data.list;
+    historyInDB.forEach(e => {
+      const key = `${e.txHash}-${e.nftId}-${e.event}`;
+      if (eventMap.has(key)) {
+        eventMap.get(key).price = e.price;
+      }
+    });
```
williamchong, line 343: it seems this function does not return/output anything?

williamchong (changes requested):

src/util/nft.js
WeiJun0827: It doesn't return because it intends to mutate the `history`  passed into the function.

williamchong: it did not modify history either, or did i miss anything?

WeiJun0827: BTW, both `populatePurchasePriceFromChain` and `populateCollectPriceFromDB` mutate `history` here.

WeiJun0827: BTW, both `populatePurchasePriceFromChain` and `populateCollectPriceFromDB` mutate `event` of `history` here.

WeiJun0827: `e` from `eventMap` is also the item in `history`, so `eventMap.get(key).price = e.price;` mutates `history` here.
(probably I made it not easy to read at first glance 🌚

williamchong: i think we should explicitly clone history the return the new modified instance to make this easier to read

WeiJun0827: Can this issue be fixed by better function naming? Since we want to `await` two async functions to mutate a single source of `history` here. If the functions return a new modified instance, the process would be sequential but not concurrent.

src/mixins/nft.js
```diff
@@ -437,37 +439,37 @@ export default {
...
       }
+
+      const eventMap = new Map();
+      history.forEach(event => {
+        eventMap.set(getEventKey(event), event);
+      });
+
+      const priceMaps = await Promise.all(promises);
+      priceMaps.forEach(priceMap => {
+        priceMap.forEach((price, key) => {
```
williamchong, line 466: So there is an hidden assumption that anything in `getCollectPriceMap` will overwrite `getCollectPriceMap` if they collide?

WeiJun0827: Yeah..., could say so. While we don't have the purchase price in our Firestore yet. I didn't check if `price` will encounter collisions for now.

src/util/nft.js
```diff
@@ -283,6 +285,59 @@ export const getNFTClassesRespectDualPrefix = async (axios, owner) => {
...
+    history.map(async e => {
+      const { event, txHash, classId, nftId } = e;
+      if (event === 'buy_nft') {
+        const price = await getPurchasePrice(axios, nftId, classId, txHash);
+        const key = getEventKey(e);
+        priceMap.set(key, price);
+      }
+    })
+  );
+  return priceMap;
```
nwingt, line 322: I don't think `Array.prototype.map` should have side effects (i.e. setting `priceMap` in this case)

```suggestion
  return new Map(
    await Promise.all(
      history.filter(e => e.event === 'buy_nft').map(async e => {
        const { txHash, classId, nftId } = e;
        const price = await getPurchasePrice(axios, nftId, classId, txHash);
        const key = getEventKey(e);
        return [key, price];
      })
    )
  );
```

src/util/nft.js
```diff
@@ -194,19 +194,21 @@ function formatNFTEvent(event) {
...
     memo,
   } = event;
   let eventName;
-  switch (event.action) {
+  switch (action) {
     case '/cosmos.nft.v1beta1.MsgSend':
       eventName =
         fromWallet === LIKECOIN_NFT_API_WALLET ? 'purchase' : 'transfer';
       break;
+    case 'buy_nft':
```
nwingt, line 207: Should ask @joshkiu for new event icon?

williamchong:
@nwingt Do you think this should target develop branch instead?

nwingt:
> @nwingt Do you think this should target develop branch instead?

To develop then

src/util/nft.js
nwingt: Confirmed with @joshkiu 
image

nwingt (approved):

PR #872, opened 2023-01-11
✨ Support displaying sell_nft message

williamchong:


nwingt (approved):

williamchong:
@nwingt Turns out I don't have to flip from to for sell events

PR #875, opened 2023-01-11
♻️ Refactor gem level logic to store

williamchong:
Fix gem level logic, should only count purchase count

### DIFF
--- a/src/components/NFTPortfolio/Base.vue
+++ b/src/components/NFTPortfolio/Base.vue
@@ -1,8 +1,5 @@
 <template>
-  <NFTGemWrapper
-    :collected-count="collectedCount"
-    :is-writing-nft="isWritingNFT"
-  >
+  <NFTGemWrapper :class-id="classId">
     <template v-slot="gem">
       <NFTPortfolioCard
         :gem-level="gem.level"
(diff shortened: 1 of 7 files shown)


src/store/modules/nft.js
```diff
@@ -30,7 +31,8 @@ const state = () => ({
 
 const mutations = {
   [TYPES.NFT_SET_NFT_CLASS_PURCHASE_INFO](state, { classId, info }) {
-    Vue.set(state.purchaseInfoByClassIdMap, classId, info);
+    const gemLevel = getGemLevelByCount(info?.metadata?.soldCount);
+    Vue.set(state.purchaseInfoByClassIdMap, classId, { gemLevel, ...info });
```
nwingt, line 35: I think should move the part `getGemLevelByCount(info?.metadata?.soldCount)` to getters `getNFTClassGemLevel`, If another mutation updates the purchase info, the gem level will be out sync

src/components/NFTPortfolio/Base.vue
```diff
@@ -1,7 +1,6 @@
 <template>
   <NFTGemWrapper
-    :collected-count="collectedCount"
-    :is-writing-nft="isWritingNFT"
+    :class-id="classId"
   >
```
nwingt, line 4: Single line?

nwingt (approved):

PR #882, opened 2023-01-19
💄 Put gem list above user description

williamchong:
Before:
image

After:
image


nwingt (approved):

nwingt:
Should merge into master?

williamchong:
> Should merge into master?

Seems they happen to be in sync now, maybe I just push both

PR #891, opened 2023-01-30
✨ Allow purchasing non-WNFT via marketplace listing

williamchong:
image
image
image
Pay by card is disabled until we support marketplace API via card


src/pages/nft/class/_classId/index.vue
```diff
@@ -77,7 +76,7 @@
                 :collector-count="ownerCount"
               />
             </template>
-            <template v-else-if="isShowPriceSection">
+            <template v-if="isShowPriceSection">
```
nwingt, line 79: I think `<NFTPagePriceSection/>` should now move b4 `<NFTPagePrimitiveClassInfoSection>`

nwingt (changes requested):

nwingt (approved):

PR #894, opened 2023-01-31
🐛 Fix undefined nftId when selling, add analytics events

williamchong:


src/pages/nft/class/_classId/index.vue
```diff
@@ -342,6 +343,15 @@ export default {
       );
       return this.handleCollect();
     },
+    handleClickSellFromControlBar() {
+      logTrackerEvent(
+        this,
+        'NFT',
+        'NFTSellClick(ClassDetailsPageControlBar)',
```
nwingt, line 350:
```suggestion
        'nft_class_details_sell_from_control_bar',
```

src/pages/nft/class/_classId/_nftId.vue
```diff
@@ -557,6 +559,15 @@ export default {
       );
       return this.handleCollect();
     },
+    handleClickSellFromControlBar() {
+      logTrackerEvent(
+        this,
+        'NFT',
+        'nft_details_click_sell_from_control_bar',
```
nwingt, line 566:
```suggestion
        'nft_details_sell_from_control_bar',
```

nwingt (approved):

PR #896, opened 2023-02-01
✨ Add mint nft button in menu and footer

williamchong:
image
image


src/components/MenuIcon.vue
```diff
@@ -3,6 +3,7 @@
   <IconLikeEvent v-else-if="type === 'civic'" />
   <IconSetting v-else-if="type === 'setting'" />
   <IconLogOut v-else-if="type === 'signOut'" />
+  <IconMint v-else-if="type === 'mintNft'" />
```
nwingt, line 6: The size of the icon should be 20x20px but `<IconMint/>` is 15px
```suggestion
  <IconMint v-else-if="type === 'mintNft'" class="w-[20px]"/>
```

PR #900, opened 2023-02-02
💄 [notification/] Update Notification page UI

AuroraHuang22:
網頁擷取_2-2-2023_191553_localhost


nwingt:
First impression:

1. The date label seemed clickable, I suggest removing the bg, smaller font and lighter color
image

2. I think the time label looks better if align to the top when there is a message and a long description, I suggest to move the label under the message or descriotion like Facebook
image
image

image

3. I think the page title can be bold,
4. also the separator is too outstanding, can use more subtle gray.
image

5. Do you want to add an inset around the list item? It seems there is a [margin + 100% width bug](https://stackoverflow.com/questions/9877379/margin-right-broken-on-width-100)
image

AuroraHuang22:
截圖 2023-02-03 下午7 02 56

src/components/SiteHeader.vue
```diff
@@ -113,6 +108,23 @@
...
+                :class="[
+                  'flex',
+                  'justify-center',
+                  'items-center',
+                  'bg-shade-gray',
+                  {'bg-danger':getNotificationCount},
+                  'text-white',
+                  'text-[6px]',
+                  'rounded-[50%]',
+                  'w-[24px]',
```
nwingt, line 123: Is the fixed width enough for the text contained? I think dynamic size is better, use `rounded-full` in that case

src/pages/notification/index.vue
```diff
@@ -0,0 +1,312 @@
...
+          ]"
+        >
+          {{ group.date }}
+        </div>
+        <!-- events -->
+        <LinkV2
+          v-for="event in group.events"
+          :key="[event.tx_hash, event.class_id, event.nft_id, event.eventType].join('-')"
+          :to="event.targetRoute"
+          :is-inline="true"
```
nwingt, line 45:
```suggestion
```

src/pages/notification/index.vue
```diff
@@ -0,0 +1,312 @@
...
+
+      events.forEach(event => {
+        let date;
+        const today = new Date().setHours(0, 0, 0, 0);
+        const yesterday = today - 86400000;
+        const eventDate = new Date(event.timestamp).setHours(0, 0, 0, 0);
+        if (eventDate === today) {
+          date = 'Today';
+        } else if (eventDate === yesterday) {
+          date = 'Yesterday';
```
nwingt, line 280: Should localize?

AuroraHuang22:
截圖 2023-02-04 下午1 47 04

src/components/SiteHeader.vue
```diff
@@ -113,6 +108,25 @@
...
             </template>
+            <template #label-append>
+              <div
+                v-if="item.value === 'notification'"
+                :class="[
+                  'flex',
+                  'justify-center',
+                  'items-center',
+                  'bg-shade-gray',
+                  {'bg-danger':getNotificationCount},
```
nwingt, line 119:
```suggestion
                  { 'bg-danger': getNotificationCount },
```

src/components/SiteHeader.vue
```diff
@@ -113,6 +108,25 @@
...
+                  'rounded-full',
+                  'min-w-[24px]',
+                  'p-[4px]'
```
nwingt, line 122:
```suggestion
                  'px-[8px]',
                  'py-[4px]',
```

AuroraHuang22:
https://user-images.githubusercontent.com/75730405/217227037-8a3810fd-8edd-4d69-97f9-12f14c092a67.mp4

src/store/modules/wallet.js
```diff
@@ -344,6 +355,7 @@ const actions = {
           e.timestamp = new Date(e.timestamp);
           return e;
         })
+        .filter(e => e.action === '/cosmos.nft.v1beta1.MsgSend')
```
williamchong, line 358: What are the other events that exists other than `/cosmos.nft.v1beta1.MsgSend`? we filtered `action_type` in API already

src/store/modules/wallet.js
```diff
@@ -152,10 +153,20 @@ const getters = {
...
     new Date(state.events[0]?.timestamp).getTime(),
+  getEventLastSeenTs: state => state.eventLastSeenTs,
   getHasUnseenEvents: state =>
     state.eventLastSeenTs &&
     state.events[0]?.timestamp &&
     state.eventLastSeenTs < new Date(state.events[0]?.timestamp).getTime(),
+  getNotificationCount: (state, getters) => {
+    const count = state.events.findIndex(
+      e => state.eventLastSeenTs > new Date(e.timestamp).getTime()
+    );
```
williamchong, line 164: We need to filter away our own purchase event from notification
a good filter would be sender = NFT API and receiver = user self wallet
Maybe should refactor and add `eventType` into store instead of in UI component

src/store/modules/wallet.js
```diff
@@ -105,7 +106,7 @@ const mutations = {
       state.email = '';
       state.emailUnverified = '';
       state.followees = [];
-      state.eventLastSeenTs = 0;
+      state.eventLastSeenTs = 1;
```
williamchong, line 109: Why 1 instead of 0?

src/pages/notification/index.vue
```diff
@@ -0,0 +1,343 @@
...
+              <div
+                v-if="!event.eventHasSeen"
+                class="bg-danger rounded-[50%] w-[8px] h-[8px]"
+              />
+            </div>
+            <!-- avatar -->
+            <Identity
+              class="self-start flex-shrink-0"
+              :avatar-url="event.displayAvatar"
+              :avatar-size="38"
```
williamchong, line 103: Do we need to know if its civic liker?

src/pages/notification/index.vue
```diff
@@ -0,0 +1,343 @@
...
+      try {
+        await this.signLogin();
+      } catch {
+        // No-op
+      }
+    }
+  },
+  beforeRouteLeave(to, form, next) {
+    this.updateEventLastSeenTs();
+    next();
```
williamchong, line 284: Does this trigger on page/browser close?

williamchong (changes requested):

src/mixins/nft.js
```diff
@@ -553,6 +554,7 @@ export default {
       }
       this.lazyGetUserInfoByAddresses([...new Set(addresses)]);
       this.isHistoryInfoLoading = false;
+      this.fetchWalletEvents();
```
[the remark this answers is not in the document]
AuroraHuang22: I plan to update getEvent after collect/transfer, maybe moving fetchWalletEvent to collect/transfer would be better？

src/store/modules/wallet.js
AuroraHuang22: since `getHasUnseenEvents: state =>
    state.eventLastSeenTs && ....`  may return false if eventLastSeenTs = 0

src/mixins/nft.js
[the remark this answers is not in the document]
AuroraHuang22: But if we don't want to receive our own purchase/transfer notifications, then I will just remove fetchWalletEvent directly.

src/pages/notification/index.vue
AuroraHuang22: well ~ It doesn‘t 🙃, I changed it to use `beforeDestroy` to make sure it will be triggered when leaving/closing the page in any situation.

src/store/modules/wallet.js
nwingt: How about -1

AuroraHuang22:
截圖 2023-02-08 下午2 28 23

src/pages/notifications/index.vue
```diff
@@ -0,0 +1,375 @@
...
+      }
+    }
+    window.addEventListener('beforeunload', () => {
+      this.updateEventLastSeenTs();
+    });
+  },
+  beforeDestroy() {
+    window.removeEventListener('beforeunload', () => {
+      this.updateEventLastSeenTs();
+    });
```
williamchong, line 309: You need to use the same function for `addEventListener` and `removeEventListener`, in which `() => {}` is always a new function and not the same one.
Also does SPA navigation (i.e. clicking `nuxt-link` or `this.$router.push()`) trigger `beforeunload` ?

williamchong (changes requested):

AuroraHuang22:
截圖 2023-02-10 上午11 39 38

src/pages/notifications/index.vue
```diff
@@ -299,14 +299,16 @@ export default {
...
-    });
+    window.addEventListener('beforeunload', this.updateEventLastSeenTs);
   },
+  // For SPA navigation
+  beforeRouteLeave(to, from, next) {
+    this.updateEventLastSeenTs();
+    next();
+  },
+  // For closing tab/browser
   beforeDestroy() {
```
williamchong, line 312: Does route leave trigger `beforeDestroy()` too? other wise `beforeRouteLeave()` needs to remove `beforeunload` handler too

PR #913, opened 2023-02-07
🚧 Add notification settings page

nwingt:
image


src/pages/settings/notification.vue
```diff
@@ -0,0 +1,38 @@
...
+            </template>
+          </SettingsListItem>
+        </li>
+        <li>
+          <SettingsListItem :is-clickable="false">
+            {{ $t('settings_notification_tx_purchase') }}
+            <template #accessory>
+              <CheckBox v-model="isPurchaseNotificationEnabled"/>
+            </template>
+          </SettingsListItem>
```
williamchong, line 21: Can we set a threshold for purchase event to only notify if sales price is larger than a certain amount?

src/pages/settings/index.vue
```diff
@@ -16,6 +16,11 @@
           <span class="settings-menu__item-title"> {{ $t('settings_following') }}</span>
         </NuxtLink>
       </li>
+      <li>
+        <NuxtLink class="settings-menu__item" :to="{ name: 'settings-notification' }">
+          <span class="settings-menu__item-title"> {{ $t('settings_notification') }}</span>
+        </NuxtLink>
+      </li>
```
williamchong, line 23: Can merge with `email` page, setting email and notification seems highly related

src/pages/settings/notification.vue
```diff
@@ -12,27 +12,73 @@
...
+  computed: {
+    ...mapGetters(['loginAddress', 'walletHasLoggedIn']),
+  },
+  watch: {
+    isTransferNotificationEnabled() {
+      if (this.isReady) {
+        this.handleToggle();
+      }
+    },
+  },
```
williamchong, line 50: Can we use `@change` on `CheckBox` instead of watch?

src/pages/settings/notification.vue
```diff
@@ -12,27 +12,73 @@
...
+      try {
+        await this.signLogin();
+      } catch {
+        // No-op
+      }
+      if (!this.walletHasLoggedIn) {
+        return;
+      }
+    }
+    await this.fetchNotification();
```
williamchong, line 62: very confusing name, at least `fetchNotificationSettings` ?

src/pages/settings/notification.vue
```diff
@@ -12,27 +12,73 @@
...
+    this.isReady = true;
+  },
+  methods: {
+    ...mapActions(['signLogin']),
+    async fetchNotification() {
+      const { data } = await this.$api.get(
+        getUserNotification(this.loginAddress)
```
williamchong, line 69: similarly `getUserNotificationSettings`

src/server/api/routes/users/v2/notification.js
```diff
@@ -0,0 +1,65 @@
...
+    try {
+      const { wallet: user } = req.params;
+      const userDoc = await walletUserCollection.doc(user).get();
+      const { notification = {} } = userDoc.data();
+      const { transfer = false } = notification;
+      res.json({
+        notification: {
+          transfer,
+        },
+      });
```
williamchong, line 27: Just return `notification` object? preferably with a filter to filter out secret (if we do have)

src/pages/settings/notification.vue
```diff
@@ -12,27 +12,73 @@
...
+      this.isTransferNotificationEnabled = transfer;
+    },
+    handleToggle() {
+      this.$api.post(getUserNotification(this.loginAddress), {
+        notification: {
+          transfer: this.isTransferNotificationEnabled,
+        },
+      });
+    },
+  },
```
williamchong, line 82: prefer generic function for this whole page, maybe `transfer` and `this.isTransferNotificationEnabled` as a input param when defining @change handler to call this function

williamchong (changes requested):

src/pages/settings/notification.vue
williamchong: Also consider putting this into `wallet` store

src/server/api/routes/users/v2/notification.js
WeiJun0827: Do we need to gerantee the fields in `notification` for this API? Or just let front-end decide the action if it's `undefined`?

williamchong: yeah having defaults here could be useful, lets keep it then

src/pages/settings/notification.vue
WeiJun0827: Seems `@input` event doesn't accept params, temporarily keep it.

PR #915, opened 2023-02-08
✨ Add top fan-ed creator list in dashboard

williamchong:
image

### DIFF
--- a/src/components/NFTPortfolio/TopUsersList.vue
+++ b/src/components/NFTPortfolio/TopUsersList.vue
@@ -1,16 +1,11 @@
 <template>
-  <CardV2
+  <component
+    :is="isCard ? 'CardV2' : 'div'"
     class="w-full"
     :has-content-padding="true"
     :is-narrow="true"
   >
-    <Label
-      class="w-min font-600 text-like-green"
-      :text="label"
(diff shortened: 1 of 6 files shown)


src/locales/en.json
```diff
@@ -1,5 +1,5 @@
 {
-  "about_nft_page_description": "Publish as you write. Collect as you read.",
+   "about_nft_page_description": "Publish as you write. Collect as you read.",
```
nwingt, line 2:
```suggestion
  "about_nft_page_description": "Publish as you write. Collect as you read.",

```

src/components/NFTPortfolio/TopUsersList.vue
```diff
@@ -1,16 +1,11 @@
...
     :is-narrow="true"
   >
-    <Label
-      class="w-min font-600 text-like-green"
-      :text="label"
-      preset="h5"
-      align="center"
-      valign="middle"
-    />
+    <slot name="label" />
```
nwingt, line 8: Seems better
```suggestion
    <slot name="prepend" />
```

nwingt (approved):

AuroraHuang22 (approved):

PR #916, opened 2023-02-08
🐛 Fix memo not reset after collect

AuroraHuang22:


src/components/EventModal/Collect/index.vue
```diff
@@ -254,6 +254,7 @@ export default {
         // Reset state when open modal
         this.resetState();
       }
+      this.memo = '';
```
nwingt, line 257: put inside `resetState()` perhaps?

src/components/EventModal/Collect/index.vue
```diff
@@ -285,6 +285,7 @@ export default {
       this.userCollectedCount = undefined;
       this.fetchNFTPrices(this.classId);
       this.fetchUserCollectedCount();
+      this.memo = '';
```
nwingt, line 288: Should put above `// Mixin`

nwingt (approved):

PR #924, opened 2023-02-09
Improve email verification UX & add events

nwingt:


williamchong (approved):
i think we need a follow up story though
verify email via token should not require login

nwingt:
> i think we need a follow up story though verify email via token should not require login

Yes, I think so.

PR #925, opened 2023-02-09
🐛 [Email] Fix possible double sending mint notification

nwingt:


williamchong (approved):
proper fix should be auto claim all subscription id with same email on user register/email verification

nwingt:
Yes, should handle that in another story

Issue #932, opened 2023-02-13
[查看精選作品] button not clickable


ckxpress:
https://liker.land/writing-nft/about

williamchong:
> https://liker.land/writing-nft/about

it's clickable on my side, can you try again, we did just have some js breaking bug which is fixed
if not, please provide a screenshot, device you are on, and if there is any console error

ckxpress:
did you try Chinese interface? for me, it's clickable on eng interface but not chi

williamchong:
> did you try Chinese interface? for me, it's clickable on eng interface but not chi

Cant reproduce the issue on both language, what device, os and browser version are you on?

ckxpress:
for me - M1, Mac OS 13.1 (22C65) - [查看精選作品] only works when i use Safari, but not Chrome/Brave/Firefox

williamchong:
@ckxpress please check if it's fixed, thanks

ckxpress:
yes thanks.

PR #933, opened 2023-02-13
🐛 Fix NFT about page CTA is blocked

williamchong:
image

The svg animation is larger in zh version, which blocks the whole CTA button
It's still partly blocked in this PR, not sure how to style it better

Related: https://github.com/likecoin/liker-land/issues/932


nwingt (commented):
Should add `pointer-events-none` class to the blocking `<svg>`  instead of adding spacing to the CTA button

AuroraHuang22 (approved):

nwingt (approved):

PR #938, opened 2023-02-16
✨ [Trending] Limit number of work from same owner

WeiJun0827:


src/pages/writing-nft/index.vue
```diff
@@ -213,17 +216,49 @@ export default {
...
+        return data.iscn_owner;
+      } catch (err) {
+        console.error(`Failed to fetch uri ${uri}`);
+        return null;
+      }
+    },
+    async filterNFTClassesByOwner(nftClasses) {
+      const ownerList = await Promise.all(
+        nftClasses
+          .slice(0, NFT_CLASS_DISPLAY_COUNT * 3)
```
nwingt, line 242: Why times 3? Should it be `MAX_NFT_CLASS_COUNT_PER_OWNER`?

src/pages/writing-nft/index.vue
```diff
@@ -213,17 +216,49 @@ export default {
...
+          .map(c => this.getOwnerFromUri(c.uri))
```
williamchong, line 243: If no special reason, owners should get from chain api instead? chain the the source of truth in this case

src/pages/writing-nft/index.vue
```diff
@@ -213,17 +216,49 @@ export default {
...
+      );
+      const filteredNFTClasses = [];
+      const ownerToNFTClassCountMap = {};
+      for (let i = 0; i < ownerList.length; i += 1) {
+        const owner = ownerList[i];
+        // eslint-disable-next-line no-continue
+        if (!owner) continue;
```
williamchong, line 250: This case should not exists, but for NFT without owner we should just show it without filtering? Since we just want to handle the case of many entries from same owner

williamchong (changes requested):

src/pages/writing-nft/index.vue
WeiJun0827: It's just an arbitrary estimated number. Since even we use `NFT_CLASS_DISPLAY_COUNT * MAX_NFT_CLASS_COUNT_PER_OWNER`, half of the works are probably from 2 or 3 owners.

nwingt: Maybe add a comment here

WeiJun0827: Added

nwingt (approved):

williamchong (approved):

PR #939, opened 2023-02-16
🐛 Disable uncollectable in latest WNFT

nwingt:
Redirect uncollectable NFT to the class details page

Before
Screenshot 2023-02-16 at 14 11 57

After
Screenshot 2023-02-16 at 14 11 49

### DIFF
--- a/src/mixins/nft.js
+++ b/src/mixins/nft.js
@@ -339,6 +339,17 @@ export default {
         ? this.listingInfo.nftId
         : this.purchaseInfo?.nftId;
     },
+    nftCollectRoute() {
+      return this.nftIsCollectable && this.nftIdCollectNext
+        ? {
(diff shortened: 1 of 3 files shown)


src/mixins/nft.js
```diff
@@ -339,6 +339,17 @@ export default {
         ? this.listingInfo.nftId
         : this.purchaseInfo?.nftId;
     },
+    nftCollectRoute() {
+      return this.nftIsCollectable
+        ? {
+            name: 'nft-class-classId-nftId',
+            params: { classId: this.classId, nftId: this.nftIdCollectNext },
+          }
```
williamchong, line 347: Should guard case where `nftIdCollectNext` not exists as well

williamchong (changes requested):

williamchong (approved):

PR #942, opened 2023-02-16
🚸 Remove cooldown for changing email if not have one

nwingt:


williamchong (approved):
Maybe we can have a large and small cooldown later

PR #943, opened 2023-02-16
🐛 [Notification] Fix last seen not update when closing window

AuroraHuang22:


src/pages/notifications/index.vue
```diff
@@ -193,11 +196,12 @@ export default {
   },
   mixins: [walletMixin],
   data() {
-    return { isLoading: false };
+    return { postTimeStamp: undefined };
```
nwingt, line 199:
```suggestion
    return { lastEntryTs: undefined };
```

nwingt (approved):

src/pages/notifications/index.vue
```diff
@@ -301,26 +305,28 @@ export default {
...
+    this.$api.$post(updateEventLastSeen());
+    this.lastEntryTs = Date.now();
     window.addEventListener('beforeunload', this.updateEventLastSeenTs);
   },
   // For SPA navigation
   beforeRouteLeave(to, from, next) {
-    this.updateEventLastSeenTs();
-    window.removeEventListener('beforeunload', this.updateEventLastSeenTs);
+    this.handleRouteLeave();
+    window.removeEventListener('beforeunload', this.handleRouteLeave);
```
williamchong, line 315: Since event list won't update unless you press refresh, `updateEventLastSeenTs` only needs to be called on mounted and on refresh? Otherwise on leave it might cause unseen even in between treated as seen

williamchong (changes requested):

src/pages/notifications/index.vue
```diff
@@ -301,26 +305,19 @@ export default {
...
     }
-    window.addEventListener('beforeunload', this.updateEventLastSeenTs);
+    this.$api.$post(updateEventLastSeen());
+    this.lastEntryTs = Date.now();
   },
   // For SPA navigation
   beforeRouteLeave(to, from, next) {
-    this.updateEventLastSeenTs();
-    window.removeEventListener('beforeunload', this.updateEventLastSeenTs);
+    this.updateEventLastSeenTs(this.lastEntryTs || Date.now());
```
williamchong, line 313: Just call `this.updateEventLastSeenTs(this.lastEntryTs)` on `mounted` or maybe in `handleRefresh `?

williamchong (changes requested):

src/pages/notifications/index.vue
williamchong: Also we should call `this.$api.$post(updateEventLastSeen());` in `handleRefresh`
maybe refactor 
```
this.updateEventLastSeenTs(Date.now());
this.$api.$post(updateEventLastSeen());
```
into a method

src/pages/notifications/index.vue
```diff
@@ -306,18 +306,23 @@ export default {
...
+    }
     next();
   },
 
   methods: {
     ...mapActions(['fetchWalletEvents', 'updateEventLastSeenTs']),
     async handleRefresh() {
+      this.updateEventLastSeenTs(this.lastUpdatedTime);
       await this.fetchWalletEvents();
+      this.$api.$post(updateEventLastSeen());
```
williamchong, line 324: can we check if there's actually new wallet event? if not we don't need to call update

williamchong (changes requested):

williamchong (approved):

PR #944, opened 2023-02-17
✨ Add email notification for transfer & purchase

nwingt:
Todos: 
- [x] Add setting in setting page
- [x] Confirm setting name in firestore
- [x] Refine email template
- [x] Add unsubscribe link

### Notification settings
image

### Transfer notification email
image

### Purchase notification email
image


src/locales/zh-Hant.json
```diff
@@ -469,6 +469,15 @@
   "settings_following_login_in_button": "驗證",
   "settings_following_login_label": "驗證錢包以查看追蹤列表",
   "settings_liker_id": "Liker ID",
+  "settings_notification": "通知",
+  "settings_notification_fetch_error_message": "載入通知設定時出現錯誤：{error}",
+  "settings_notification_title": "通知設定",
+  "settings_notification_tx_purchase": "收藏",
+  "settings_notification_tx_purchase_price_threshold": "當價格大於上述設定值時便會發出通知",
```
williamchong, line 476:
```suggestion
  "settings_notification_tx_purchase_price_threshold": "當交易價格大於上述設定值時便會發出通知",

```

src/components/CheckBox.vue
```diff
@@ -0,0 +1,58 @@
...
+      ]"
+    >
+      <input
+        :class="[
+          'absolute',
+          'w-0',
+          'h-0',
+          'opacity-0',
+          'inset',
+        ]"
```
williamchong, line 17: Passing `disabled` might be useful?

src/server/api/middleware/auth.js
```diff
@@ -7,6 +9,33 @@ function authenticateV2Login(req, res, next) {
...
+    res.status(400).send('ADDRESS_MISMATCH');
+    return;
+  }
+  next();
+}
+
+async function checkEmailHasVerified(req, res, next) {
+  const { wallet } = req.params;
+  const userDoc = await walletUserCollection.doc(wallet).get();
+  const { email, emailUnconfirmed } = userDoc.data();
```
williamchong, line 25: Do we need to guard `userDoc.data()` not exists?

src/server/api/routes/users/v2/notification.js
```diff
@@ -0,0 +1,64 @@
...
+  }
+});
+
+router.post(
+  '/notification',
+  authenticateV2Login,
+  checkEmailHasVerified,
+  async (req, res, next) => {
+    try {
+      setPrivateCacheHeader(res);
```
williamchong, line 35: We should set private cache header in `authenticateV2Login`? Both security and code styling sense it seems to be an improvement
Also setting private cache for POST seems redundant

src/server/api/routes/users/v2/notification.js
```diff
@@ -0,0 +1,64 @@
...
+const { walletUserCollection } = require('../../../../modules/firebase');
+
+const router = Router();
+
+router.get('/notification', authenticateV2Login, async (req, res, next) => {
+  try {
+    const { user } = req.session;
+    const userDoc = await walletUserCollection.doc(user).get();
+    const { notification = {} } = userDoc.data();
+    const { transfer = true, purchasePrice = 0 } = notification;
```
williamchong, line 17: Should we call `setPrivateCacheHeader` here instead?

src/components/CheckBox.vue
williamchong: Seems `isDisabled` is used but not implemented

functions/pubsub/handlers.js
```diff
@@ -0,0 +1,271 @@
...
+
+  const { name, image } = await fetchNFTMetadata(classId);
+
+  const userDoc = await walletUserCollection.doc(toAddress).get();
+  const userDocData = userDoc.data();
+  if (!userDocData) return;
+
+  const {
+    email,
+    notification: { transfer = false } = {},
```
williamchong, line 227: transfer is default to `true` in GET api, but `false` here, potential inconsistency? Maybe should be `true` here also?

williamchong (changes requested):

src/server/api/middleware/auth.js
WeiJun0827: Logically we should have created a doc for a logged-in user, else the request would be blocked by the `authenticateV2Login` middleware before calling `checkEmailHasVerified`?

WeiJun0827: But that's the case if we always check `authenticateV2Login` first... 🌚

williamchong (approved):

williamchong (approved):

PR #946, opened 2023-02-17
🌐 Add language setting API

WeiJun0827:


src/server/api/routes/users/v2/locale.js
```diff
@@ -0,0 +1,44 @@
...
+const { authenticateV2Login } = require('../../../middleware/auth');
+const { setPrivateCacheHeader } = require('../../../middleware/cache');
+const { handleRestfulError } = require('../../../middleware/error');
+const { walletUserCollection } = require('../../../../modules/firebase');
+
+const { DEFAULT_LOCALE, AVAILABLE_LOCALES } = require('../../../constant');
+
+const router = Router();
+
+router.get('/locale', authenticateV2Login, async (req, res, next) => {
```
WeiJun0827, line 11: Not sure if we need this API though

src/server/api/constant/index.js
```diff
@@ -48,6 +48,9 @@ const CIVIC_LIKER_QUANTITY_OPTIONS = [1];
 
 const VERIFICATION_EMAIL_RESEND_COOLDOWN_IN_MS = IS_TESTNET ? 60000 : 3600000;
 
+const DEFAULT_LOCALE = 'en';
+const AVAILABLE_LOCALES = ['en', 'zh-Hant'];
```
nwingt, line 52: can this read from nuxt.config.js?

src/server/api/routes/users/v2/locale.js
nwingt: Keep it, maybe useful

nwingt (commented):
I think we need to set locale when login too

src/store/modules/wallet.js
```diff
@@ -404,11 +404,12 @@ const actions = {
       commit(WALLET_SET_LIKE_BALANCE_FETCH_PROMISE, undefined);
     }
   },
-  async walletFetchSessionUserInfo({ commit }, address) {
+  async walletFetchSessionUserInfo({ commit, dispatch }, address) {
     try {
       commit(WALLET_SET_FOLLOWEES_FETCHING_STATE, true);
       const userInfo = await this.$api.$get(getUserV2Self());
       commit(WALLET_SET_USER_INFO, userInfo || { user: address });
+      dispatch('setLocale', userInfo.locale);
```
williamchong, line 412: Not sure if we need to check if `userInfo.locale` exists and is a valid locale in `setLocale`

williamchong (approved):
I was imagining the check to be inside serLocale but this would work too

williamchong (approved):

PR #950, opened 2023-02-20
🚸 Do not trigger login if window is not focused

nwingt:


williamchong (commented):
Can we check once we are in focus and trigger again?

nwingt:
Shall follow up UX in another story

PR #952, opened 2023-02-21
🚸 Update last seen ts with latestEventTimestamp

AuroraHuang22:


src/pages/notifications/index.vue
```diff
@@ -206,7 +206,11 @@ export default {
       'getNFTClassMetadataById',
       'getNotificationCount',
       'getEventLastSeenTs',
+      'getLatestEventTimestamp',
     ]),
+    latestEventTimestamp() {
+      return this.getLatestEventTimestamp;
+    },
```
nwingt, line 213: Why computed??

AuroraHuang22: I thought the **immediate: true** property would only start watching after the **created** lifecycle hook, so I was worried that if `getLatestEventTimestamp` already has a value before entering the page, the handler wouldn't be triggered. That's why I wrapped it in a computed property.

AuroraHuang22: But later I found that if immediate: true, the handler will be triggered immediately without going through a process of going from undefined to having a value. 😝

src/server/api/routes/users/v2/event.js
```diff
@@ -1,20 +1,41 @@
...
 } = require('../../../../modules/firebase');
 const { authenticateV2Login } = require('../../../middleware/auth');
 const { setPrivateCacheHeader } = require('../../../middleware/cache');
 const { publisher, PUBSUB_TOPIC_MISC } = require('../../../../modules/pubsub');
 
 const router = Router();
 
+function isValidTs(ts, lastSeenTs) {
+  // eslint-disable-next-line no-restricted-globals
+  if (isNaN(ts) || ts % 1000 !== 0) {
```
williamchong, line 14: `Number.isNaN` is allowed without disabling eslint, do you really need `isNaN` but not `Number.isNaN`?

src/server/api/routes/users/v2/event.js
```diff
@@ -1,20 +1,41 @@
...
+    return false;
+  }
+  if (ts < lastSeenTs) {
+    return false;
+  }
+  if (ts > Date.now() + 5 * 60 * 1000) {
+    return false;
+  }
+  return true;
```
williamchong, line 23: I think these condition can be converted into a multi line boolean, in which we can just return its value
i.e.
```
return (!(isNaN(ts) || ts % 1000 !== 0)
  && (ts > lastSeenTs)
  && (ts < Date.now() + 5 * 60 * 1000)
```

src/pages/notifications/index.vue
```diff
@@ -297,6 +298,23 @@ export default {
...
+  watch: {
+    getLatestEventTimestamp: {
+      immediate: true,
+      handler(newVal) {
+        if (newVal) {
+          try {
+            this.$api.$post(updateEventLastSeen(), {
+              ts: Number(newVal),
+            });
+            this.lastUpdatedTime = newVal;
```
williamchong, line 310: Does this trigger if I don't refresh the events at all? i.e. when is `this.$api.$post(updateEventLastSeen()` triggered if i don't refresh?

src/server/api/routes/users/v2/event.js
```diff
@@ -1,20 +1,41 @@
...
+    const userDoc = await walletUserCollection.doc(user).get();
+    const { eventLastSeenTs } = userDoc.data();
+    if (!isValidTs(ts, eventLastSeenTs)) {
+      res.status(400).send('INVALID_TIMESTAMP_PROVIDED');
+      return;
+    }
     await walletUserCollection.doc(user).update({
-      eventLastSeenTs: FieldValue.serverTimestamp(),
+      eventLastSeenTs: Timestamp.fromMillis(Math.floor(ts / 1000) * 1000),
     });
```
williamchong, line 39: Note that this `get()` then` update()` check normally requires a transaction, but it kind of doesn't matter here

src/server/api/routes/users/v2/event.js
```diff
@@ -10,33 +11,31 @@ const { publisher, PUBSUB_TOPIC_MISC } = require('../../../../modules/pubsub');
...
-      eventLastSeenTs: Timestamp.fromMillis(Math.floor(ts / 1000) * 1000),
+    await db.runTransaction(async t => {
+      const userRef = walletUserCollection.doc(user);
+      const userDoc = await t.get(userRef);
+      const { eventLastSeenTs = 1000 } = userDoc.data();
+      if (!isValidTs(ts, eventLastSeenTs)) {
+        res.status(400).send('INVALID_TIMESTAMP_PROVIDED');
+        return;
+      }
+      await t.update(userRef, {
```
williamchong, line 34: t.update is not a promise, no need await

src/server/api/routes/users/v2/event.js
```diff
@@ -10,33 +11,31 @@ const { publisher, PUBSUB_TOPIC_MISC } = require('../../../../modules/pubsub');
...
-      return;
-    }
-    await walletUserCollection.doc(user).update({
-      eventLastSeenTs: Timestamp.fromMillis(Math.floor(ts / 1000) * 1000),
+    await db.runTransaction(async t => {
+      const userRef = walletUserCollection.doc(user);
+      const userDoc = await t.get(userRef);
+      const { eventLastSeenTs = 1000 } = userDoc.data();
+      if (!isValidTs(ts, eventLastSeenTs)) {
+        res.status(400).send('INVALID_TIMESTAMP_PROVIDED');
```
williamchong, line 31: Should throw an error here to quit the transaction, then handle the res.send() part in an error handler
We want to keep everything inside a transaction simple

williamchong (changes requested):

williamchong (approved):

PR #954, opened 2023-02-21
✨ Buy listing NFT with credit card

WeiJun0827:
Depends on https://github.com/likecoin/likecoin-api-public/pull/432

### DIFF
--- a/src/mixins/nft.js
+++ b/src/mixins/nft.js
@@ -774,12 +774,17 @@ export default {
     },
     async collectNFTWithStripe({ memo = '' }) {
       try {
+        const body = { memo };
+        if (this.nftIsUseListingPrice) {
+          body.nftId = this.listingInfo.nftId;
+          body.seller = this.listingInfo.seller;
+        }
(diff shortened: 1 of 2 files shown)


src/mixins/nft.js
```diff
@@ -477,10 +479,14 @@ export default {
...
       try {
-        const { fiatPrice } = await this.$axios.$get(
+        const { fiatPrice, isListing, listingInfo } = await this.$axios.$get(
           getStripeFiatPrice({ classId })
         );
         this.nftPriceInUSD = fiatPrice;
+        if (isListing) {
+          this.listingNftId = listingInfo.nftId;
+          this.listingNftSeller = listingInfo.seller;
+        }
```
williamchong, line 489: Can we make use of `nftIsUseListingPrice` and `listingInfo` instead of making duplicate new fields? Or at least we should refactor current useListing logic when buying with LIKE into the new fields

williamchong (changes requested):

williamchong (approved):

PR #958, opened 2023-02-22
🐛 Fetch followees from separated API

nwingt:


src/pages/settings/following.vue
```diff
@@ -104,6 +104,23 @@ export default {
       }));
     },
   },
+  watch: {
+    walletHasLoggedIn: {
+      immediate: true,
+      async handler(hasLoggedIn) {
+        // Only fetch followees in client side
+        if (process.client && hasLoggedIn) {
```
williamchong, line 112: Does this cause ssr outsync error? can we just call this in mounted?

nwingt: I don't see any SSR out sync

nwingt: image

https://user-images.githubusercontent.com/17264716/220703471-0db037a7-002c-4899-a6ba-e1c00eefe397.mov

williamchong (approved):

PR #964, opened 2023-02-24
♻️ Refactor routes and correct `/followees` routes

WeiJun0827:


WeiJun0827:
Note that users received a verification email for his/her email before deploying this PR but clicking the confirmation link later would encounter an error.
But if users ask to resend another verification email should fix this case.

src/server/api/routes/users/v2/email.js
```diff
@@ -21,75 +18,70 @@ const { EXTERNAL_URL } = require('../../../../config/config');
...
-        return;
+router.post('/email', authenticateV2Login, async (req, res, next) => {
+  try {
+    const { user } = req.session;
+    const { email } = req.query;
+    if (!email) {
+      res.status(400).send('MISSING_EMAIL');
+      return;
+    }
+    console.log(user, email);
```
williamchong, line 29: Please remove `console.log()`

src/server/api/routes/users/v2/nfts/display-state.js
```diff
@@ -23,9 +20,9 @@ const NFT_DISPLAY_STATE = {
 
 const router = Router();
 
-router.get('/:wallet/nfts/display-state', async (req, res, next) => {
+router.get('/nfts/display-state', async (req, res, next) => {
```
williamchong, line 23: We do want to keep `/:wallet/` in public endpoints though? Since these are really under some kind of wallet without implicit session

src/server/api/routes/users/v2/nfts/display-state.js
```diff
@@ -40,13 +37,12 @@ router.get('/:wallet/nfts/display-state', async (req, res, next) => {
 });
 
 router.post(
-  '/:wallet/nfts/display-state',
+  '/nfts/display-state',
```
williamchong, line 40: Similarly maybe we should keep `/:wallet` here

williamchong (changes requested):

williamchong (approved):

PR #966, opened 2023-02-27
🐛 Fix trending tab range was not set to 14 days

williamchong:


nwingt (approved):

src/util/api/index.js
```diff
@@ -277,8 +277,16 @@ export const getStripeFiatPrice = ({ classId }) => {
...
   `${LIKECOIN_CHAIN_API}/likechain/likenft/v1/class?reverse=true`;
-export const getTopNFTClasses = () =>
-  `${LIKECOIN_CHAIN_API}/likechain/likenft/v1/ranking?ignore_list=${LIKECOIN_NFT_API_WALLET}`;
+export const getTopNFTClasses = ({ before, after }) => {
+  const qsPayload = {
+    before,
+    after,
+    ignore_list: LIKECOIN_NFT_API_WALLET,
+  };
+  return `${LIKECOIN_CHAIN_API}/likechain/likenft/v1/ranking??${querystring.stringify(
```
nnkken, line 286: Doubled `?`, redundant
Also should add default values `0` for `before` and `after` for robustness? Otherwise it's sending `?before=&after=&...` which relies on how API parses empty strings

williamchong: updated to only set keys in qs object if not undefined

src/util/api/index.js
```diff
@@ -280,10 +280,11 @@ export const getLatestNFTClasses = () =>
 export const getTopNFTClasses = ({ before, after }) => {
```
nnkken, line 280:
```suggestion
export const getTopNFTClasses = ({ before = 0, after = 0 }) => {
```

src/util/api/index.js
```diff
@@ -280,10 +280,11 @@ export const getLatestNFTClasses = () =>
...
   const qsPayload = {
     before,
-    after,
     ignore_list: LIKECOIN_NFT_API_WALLET,
   };
-  return `${LIKECOIN_CHAIN_API}/likechain/likenft/v1/ranking??${querystring.stringify(
+  if (after !== undefined) qsPayload.after = after;
+  if (before !== undefined) qsPayload.before = before;
```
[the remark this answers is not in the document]
williamchong: I actually wanted to remove both `before` and `after` from `const qsPayload `...
If they are removed from qsPayload then this will work? despite having them in function prototype

PR #967, opened 2023-02-27
🔧 Fine tune default camera angle for new model

williamchong:
Before
image
After
image


nwingt:
seems the back still facing the camera, not the front???

williamchong:
> seems the back still facing the camera, not the front???

What do you mean by front, the other side of the cover?

nwingt (approved):

PR #969, opened 2023-03-01
👽 Fetch followee API bafore getEvents

AuroraHuang22:
- Fetch followee before displaying the notification count. (signLogin)
- Re-fetch the followee list to update the event sender. (handleRefresh)

### DIFF
--- a/src/store/modules/wallet.js
+++ b/src/store/modules/wallet.js
@@ -446,7 +446,10 @@ const actions = {
         from: address,
       };
       await this.$api.post(postUserV2Login(), data);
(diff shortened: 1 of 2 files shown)


src/store/modules/wallet.js
```diff
@@ -447,6 +447,7 @@ const actions = {
       };
       await this.$api.post(postUserV2Login(), data);
       await dispatch('walletFetchSessionUserInfo', address);
+      await dispatch('walletFetchFollowees');
```
nwingt, line 450: Can be parallel

nwingt (approved):

PR #973, opened 2023-03-01
🎨 Better fetch session user data

nwingt:
image

### DIFF
--- a/src/mixins/wallet.js
+++ b/src/mixins/wallet.js
@@ -60,7 +60,7 @@ export default {
       'signLogin',
       'walletFetchFollowees',
     ]),
-    async connectWallet({ isSkippedLogin = false } = {}) {
+    async connectWallet({ shouldSkipLogin = false } = {}) {
       const connection = await this.openConnectWalletModal({
         language: this.$i18n.locale.split('-')[0],
       });
@@ -73,7 +73,7 @@ export default {
         'connected_wallet',
         1
       );
-      return isSkippedLogin
+      return shouldSkipLogin
         ? this.initWallet(connection)
         : this.initWalletAndLogin(connection);
     },
--- a/src/mixins/wallet.js
+++ b/src/mixins/wallet.js
@@ -52,7 +52,7 @@ export default {
     ...mapActions([
       'openConnectWalletModal',
       'disconnectWallet',
-      'initWallet',
+      'initWalletAndLogin',
       'initIfNecessary',
       'restoreSession',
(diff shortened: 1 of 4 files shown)


AuroraHuang22 (approved):

src/store/modules/wallet.js
```diff
@@ -230,17 +230,14 @@ const actions = {
     try {
       if (state.signer && !getters.walletIsMatchedSession) {
         await dispatch('signLogin');
+      } else if (getters.walletHasLoggedIn) {
```
williamchong, line 233: The else relationship seems not intuitive, i.e. what does `!(state.signer && !getters.walletIsMatchedSession)` mean?
Maybe should name the if condition above to a meaningful name

nwingt: Updated

williamchong: Is it possible for `!state.signer & getters.walletIsMatchedSession` to exist?

nwingt: Updated with flipped conditions

src/mixins/wallet.js
```diff
@@ -72,7 +73,9 @@ export default {
         'connected_wallet',
         1
       );
-      return this.initWallet(connection);
+      return isSkippedLogin
```
williamchong, line 76: Does this flag means `Login was skipped` or it means we want to skip login when calling this function?
If it is latter, we should call this flag `skipLogin`/`isSkipLogin`

src/store/modules/wallet.js
```diff
@@ -403,15 +410,28 @@ const actions = {
...
     try {
       const userInfo = await this.$api.$get(getUserV2Self());
-      commit(WALLET_SET_USER_INFO, userInfo || { user: address });
+      commit(WALLET_SET_USER_INFO, userInfo || { user: state.address });
       return userInfo;
     } catch (error) {
       throw error;
     }
   },
+  async walletFetchSessionUserData({ dispatch }, { isExtraOnly = false } = {}) {
```
williamchong, line 422: `isSkipUserInfo` for consistent naming?

williamchong (changes requested):

src/mixins/wallet.js
nwingt: How about `shouldSkipLogin`?

williamchong: Yes it would be nice

williamchong (approved):

PR #987, opened 2023-03-07
🐛 Fix notificationCount not including mint_nft

AuroraHuang22:


src/store/modules/wallet.js
```diff
@@ -157,14 +157,21 @@ const getters = {
...
-    return getters.getEvents.filter(
-      e =>
+    return getEvents.reduce((count, e) => {
+      if (
         state.eventLastSeenTs < new Date(e.timestamp).getTime() &&
-        (e.eventType === 'nft_sale' || e.eventType === 'receive_nft')
-    ).length;
+        (e.eventType === 'nft_sale' ||
+          e.eventType === 'receive_nft' ||
+          e.eventType === 'mint_nft')
```
nwingt, line 169: Use `Array.prototype.includes`?

nwingt (approved):

src/store/modules/wallet.js
```diff
@@ -157,14 +157,19 @@ const getters = {
...
       return 0;
     }
-    return getters.getEvents.filter(
-      e =>
+    return getEvents.reduce((count, e) => {
+      if (
         state.eventLastSeenTs < new Date(e.timestamp).getTime() &&
-        (e.eventType === 'nft_sale' || e.eventType === 'receive_nft')
-    ).length;
+        ['nft_sale', 'receive_nft', 'mint_nft'].includes(e.eventType)
```
williamchong, line 167: Aren't we using new_class event though

williamchong (approved):

src/store/modules/wallet.js
AuroraHuang22: have formatted [eventType](https://github.com/likecoin/liker-land/blob/develop/src/store/modules/wallet.js#L180-L181)

nwingt: Maybe the event type should be declared as constant for reuse

PR #997, opened 2023-03-10
🐛 Fix various warning & missing definition

nwingt:


src/components/NFTPortfolio/TopUsersList.vue
```diff
@@ -20,11 +20,11 @@
           @mouseenter.native.once="onHover(user.index)"
         >
           <LinkV2
-            :to="localeLocation({
+            :to="user.id ? localeLocation({
```
williamchong, line 23: `user.id` is the `key` for `v-for` though, should we actually skip item without `id` in data?

williamchong (approved):

WeiJun0827 (approved):

PR #998, opened 2023-03-10
🚸 Campaign creators should link to created tab

williamchong:


src/components/NFTCampaign/ItemBase.vue
```diff
@@ -55,7 +55,7 @@
           <h3 class="text-[24px] leading-[1.3] font-[700] text-like-green font-serif">{{ storyTitle }}</h3>
           <NuxtLink
             class="mt-[16px] flex items-center text-like-green group -my-[8px]"
-            :to="`/${ownerAddress}`"
+            :to="{ name: 'id', params: { id: ownerAddress }, query: { tab: 'created' } }"
```
nwingt, line 58: Need guarding `ownerAddress` like in #997

nwingt (changes requested):

nwingt (approved):

PR #1006, opened 2023-03-15
🚑 Fix collect/transfer requires login

nwingt:


williamchong (approved):
Would this cause some places that requires login to break though

nwingt:
> Would this cause some places that requires login to break though

Require login should guard like this
https://github.com/likecoin/liker-land/blob/544b456cd81a59c3ce90ed81891d76e1a6b01726/src/pages/notifications/index.vue#L332-L340

PR #1007, opened 2023-03-15
🎨 Mock NFT mint event for ISCN transfer by replacing chain API

WeiJun0827:
Note that the complexity of getting API will go from 1 to N, where N is the number of followees.


WeiJun0827:
> Note that the complexity of getting API will go from 1 to N, where N is the number of followees.

Fixed by making chain API accepts multiple `iscn_owner`.

williamchong (approved):

PR #1018, opened 2023-03-31
✨ Hide view content for nft book if not collected

williamchong:
Need help on making `ButtonV2` disabling properly, then I won't have to hack with `href` and `url`
Now it can still click through when `isDisabled` is true
image


src/components/NFTCampaign/ItemBase.vue
```diff
@@ -24,7 +24,7 @@
           />
           <a
             class="transition-colors cursor-pointer hover:text-like-cyan-dark flex items-center justify-center text-medium-gray mt-[8px]"
-            :href="url"
+            :href="isContentViewable ? url : ''"
```
nwingt, line 27: Can keep `url` after merging https://github.com/likecoin/liker-land/pull/1021

src/components/NFTPage/PreviewCard.vue
```diff
@@ -96,6 +97,26 @@
         </template>
       </ButtonV2>
 
+      <template v-if="isContentViewable">
+        <ButtonV2
+          v-for="contentUrl in contentUrls"
+          :key="contentUrl"
+          class="mt-[12px]"
+          preset="outline"
+          :text="$t(`nft_details_page_button_view_${getContentUrlType(contentUrl)}`)"
```
nwingt, line 106: preferable not to generate i18n key paths dynamically, it is problematic when removing unused paths/find & replace

src/components/NFTPage/PreviewCard.vue
```diff
@@ -68,9 +68,10 @@
       <ButtonV2
         v-if="url"
         preset="outline"
-        :text="$t('campaign_nft_item_view_details_label')"
-        :href="url"
+        :text="isContentViewable ? $t('nft_details_page_button_view') : $t('nft_details_page_button_collect_to_view')"
```
nwingt, line 71: Suggest adding `w-full` class for view buttons
Look very bad when there are multiple buttons with different widths
image

src/components/NFTPage/PreviewCard.vue
```diff
@@ -96,6 +97,26 @@
         </template>
       </ButtonV2>
 
+      <template v-if="isContentViewable">
+        <ButtonV2
+          v-for="contentUrl in contentUrls"
+          :key="contentUrl"
+          class="mt-[12px]"
```
nwingt, line 104: Suggest adding `w-full` class for view buttons
```suggestion
          class="mt-[12px] w-full"
```

nwingt (changes requested):

nwingt (approved):

PR #1023, opened 2023-04-06
🔧 Allow all WNFT to be collected with Stripe

nwingt:


src/mixins/nft.js
```diff
@@ -637,7 +628,7 @@ export default {
         };
         logPurchaseFlowEvent(this, 'add_to_cart', purchaseEventParams);
         logPurchaseFlowEvent(this, 'begin_checkout', purchaseEventParams);
-        if (!this.canCollectWithoutWallet && !this.getAddress) {
+        if (!this.getAddress) {
           const isConnected = await this.connectWallet({
             shouldSkipLogin: true,
           });
```
williamchong, line 634: Should this whole part be removed and just leave `await this.initIfNecessary();`? 
`this.canCollectWithoutWallet` is always `true` when `LIKECOIN_NFT_COLLECT_WITHOUT_WALLET_ITEMS_BY_CREATORS` is removed

williamchong (changes requested):

williamchong (approved):

PR #1038, opened 2023-04-19
⚗️ AB test for class page to collect CTA

AuroraHuang22:
**variant**
網頁擷取_19-4-2023_15398_localhost

**origin**
網頁擷取_19-4-2023_154815_localhost


src/experiments/class-collect-cta.js
```diff
@@ -0,0 +1,10 @@
+export default {
+  name: 'class-collect-cta',
+  experimentID: 'cljcOGr3RX-jUiSHCaMu-A',
+  isEligible: ({ route }) => {
+    const parts = route.name.split('___');
+    const routeName = parts[0];
```
nwingt, line 6: I think it is more robust to use `getRouteBaseName()` from context to extract the route name without localization

You may need to inject `getRouteBaseName` from [here](https://github.com/likecoin/liker-land/blob/57c8784c02c880150c75ae10f7c541cb4aa2c6b7/src/mixins/experiment.js#L14) in order to use it

src/pages/nft/class/_classId/index.vue
```diff
@@ -454,7 +423,9 @@ export default {
       logTrackerEvent(
         this,
         'NFT',
-        'NFTCollect(DetailsPageCTA)',
+        `NFTCollect(DetailsPageCTA_${
+          this.isExperimenting ? 'variant' : 'origin'
+        })`,
```
nwingt, line 428: It is better to use a full path without constructing dynamically

nwingt (changes requested):

src/mixins/experiment.js
```diff
@@ -11,7 +11,7 @@ export default function experimentsMixin(
         const { name, $activeVariants } = this.$exp;
         if (
           this.$exp.isEligible &&
-          !this.$exp.isEligible({ route: this.$route })
+          !this.$exp.isEligible({ route: this.getRouteBaseName(this.$route) })
```
nwingt, line 14: should pass `getRouteBaseName` into `isEligible()` as route is an object with query and params

src/pages/nft/class/_classId/index.vue
```diff
@@ -423,14 +423,23 @@ export default {
...
+    handleCollectFromCTA_exp() {
+      logTrackerEvent(
+        this,
+        'NFT',
+        'NFTCollect(DetailsPageCTA_exp)',
+        this.classId,
+        1
+      );
+      return this.handleCollect();
+    },
```
nwingt, line 441: I think no need to split, this should be handled with Optimize

nwingt (changes requested):

src/mixins/experiment.js
```diff
@@ -11,12 +11,18 @@ export default function experimentsMixin(
         const { name, $activeVariants } = this.$exp;
         if (
           this.$exp.isEligible &&
-          !this.$exp.isEligible({ route: this.$route })
+          !this.$exp.isEligible({
+            route: {
+              name: this.getRouteBaseName(this.$route),
+              query: this.$route.query,
+              params: this.$route.query,
+            },
```
nwingt, line 19:
```suggestion
            route: {
              name: this.getRouteBaseName(this.$route),
              query: this.$route.query,
              params: this.$route.params,
            },
```
I think this is more simple
```suggestion
            route: {
              ...this.$route,
              name: this.getRouteBaseName(this.$route),
            },
```

nwingt (approved):

src/components/NFTCover.vue
```diff
@@ -23,7 +23,7 @@
       @error="handleImageError"
     >
     <img
-      v-if="!isShowVideo && (!isShowImage || !isLoaded)"
+      v-if="(!isShowVideo && !isShowImage) || !isLoaded"
```
williamchong, line 26: Is this diff intentional?

williamchong (changes requested):

src/components/NFTCover.vue
AuroraHuang22: Oh, that was unintentional

williamchong (approved):

PR #1039, opened 2023-04-19
🚸 Add poster and fallback for <video>

williamchong:


nwingt (approved):

ckxpress:
stupid question. the fallback is only for webm but not mp4 right?

williamchong:
> stupid question. the fallback is only for webm but not mp4 right?

For any video error case

PR #1040, opened 2023-04-20
👔 Make email of wallet user be unique

WeiJun0827:


src/server/api/routes/users/v2/email.js
```diff
@@ -31,14 +31,16 @@ router.post('/email', authenticateV2Login, async (req, res, next) => {
...
-        emailUnconfirmed,
-        emailLastUpdatedTs,
-      } = userDoc.data();
-      if (email === currentEmail) {
-        throw new Error('EMAIL_ALREADY_UPDATED');
+      const [userDoc, userDocWithSameEmail] = await Promise.all([
+        t.get(userRef),
+        t.get(walletUserCollection.where('email', '==', email).limit(1)),
+      ]);
+      const { emailUnconfirmed, emailLastUpdatedTs } = userDoc.data();
```
nwingt, line 38: Seems this line can be put after the if block below

WeiJun0827: Yes, that would be better!

nwingt (approved):

PR #1044, opened 2023-04-20
✨ Add querying wallet by email API

WeiJun0827:


nwingt (approved):

src/server/api/routes/users/v2/wallet.js
```diff
@@ -0,0 +1,33 @@
+const { Router } = require('express');
+const { walletUserCollection } = require('../../../../modules/firebase');
+const { GET_WALLET_API_SECRET } = require('../../../../config/config');
+
+const router = Router();
+
+router.get('/wallet', async (req, res, next) => {
+  try {
+    const { email, key } = req.query;
```
williamchong, line 9: key should not be in query string, use HTTP header like `Authorization` or `X-LIKERLAND-API-KEY`

src/server/api/routes/users/v2/wallet.js
```diff
@@ -0,0 +1,33 @@
...
+    if (!email) {
+      res.status(400).send('MISSING_EMAIL');
+      return;
+    }
+    if (key !== GET_WALLET_API_SECRET) {
```
williamchong, line 14: We might want to make an extra guard that throws 403 on `GET_WALLET_API_SECRET` is not set, in case of deployment error that we did not set a `GET_WALLET_API_SECRET`

williamchong (changes requested):

src/server/config/config.js
```diff
@@ -19,4 +19,6 @@ config.LIKEPAY_HOOK = process.env.LIKEPAY_HOOK;
 
 config.EXTERNAL_URL = process.env.EXTERNAL_URL;
 
+config.GET_WALLET_API_SECRET = process.env.GET_WALLET_API_SECRET;
```
williamchong, line 22: Please also add this as `secrets` in `functions/api/index`

williamchong (changes requested):

PR #1051, opened 2023-04-24
✨ Add alert banner for NFT Book sale

nwingt:
image
image

### Mobile
image

### DIFF
--- a/src/layouts/default.vue
+++ b/src/layouts/default.vue
@@ -1,5 +1,23 @@
 <template>
   <div class="flex flex-col items-stretch min-h-screen">
+    <AlertBanner
+      :primary-button-text="$t('alert_banner_actions_purchase')"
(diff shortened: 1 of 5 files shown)


src/layouts/default.vue
```diff
@@ -1,5 +1,23 @@
 <template>
   <div class="flex flex-col items-stretch min-h-screen">
+    <AlertBanner
+      :primary-button-text="$t('alert_banner_actions_purchase')"
```
williamchong, line 4: Do we need v-if `$t('alert_banner_content_heavenly_creations_life_like')` or sth?

williamchong (approved):

src/layouts/default.vue
nwingt: Probably remove the whole component instead of removing the i18n key if not used

PR #1054, opened 2023-04-25
⚡️ Directly fetch time descending NFT events

WeiJun0827:


williamchong (approved):
This doesn't make the flow faster though? We are still looping through all pages

WeiJun0827:
> This doesn't make the flow faster though? We are still looping through all pages

Skip `.reverse()` and `.sort()`, not really significant.
I think we should prevent looping fetch to radically improve the performance.

PR #1056, opened 2023-04-25
🚸 Display latest events before fetching all events  

WeiJun0827:


williamchong (changes requested):

src/mixins/nft.js
```diff
@@ -517,109 +515,95 @@ export default {
...
-          }
-        });
-      }
-      history.sort((a, b) => b.timestamp - a.timestamp);
+            // add original event after grant event for time reverse order
+            historyWithGrant.push(event);
+          });
+          this.NFTHistory.push(...historyWithGrant);
+        } else {
+          this.NFTHistory.push(...history);
```
williamchong, line 571: When do we reset `this.NFTHistory`?

WeiJun0827: Probably safer to reset it before pushing. Nice caught.

williamchong: Would be easier to read if you return `...historyWithGrant` or `...history` in this function
then explicitly call push when calling i.e. `this.NFTHistory.push(addGrantEventForWNFTAndUpdateUserInfo(...))`

williamchong: Would be even cleaner if we completely avoid using `this`, and use function param instead in `addGrantEventForWNFTAndUpdateUserInfo`

WeiJun0827: updated, see if it's cleaner

williamchong (approved):

nwingt (approved):

PR #1047, opened 2023-04-21
🚧 Accept predefined email from URL

WeiJun0827:
截圖 2023-04-21 下午6 23 14
截圖 2023-04-21 下午6 23 32

Further UI improvement is welcome.
Not sure whether we should also provide some warning sentence, in case of abuse of this mode.

PR #1057, opened 2023-04-26
💬 Update invalid token error message

WeiJun0827:


WeiJun0827:
Cherry-picked commit to https://github.com/likecoin/liker-land/pull/1047

PR #1081, opened 2023-05-12
💩 Fix SSR error

nwingt:
The issue is probably within `<NFTCampaignItem>`, should investigate further

### DIFF
--- a/src/components/NFTSupply/Row.vue
+++ b/src/components/NFTSupply/Row.vue
@@ -7,7 +7,7 @@
       }
     ]"
   >
-    <td v-if="!isDisabled" class="w-[64px]">
+    <td v-show="!isDisabled" class="w-[64px]">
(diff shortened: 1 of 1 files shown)


src/pages/writing-nft/index.vue
```diff
@@ -73,20 +73,22 @@
...
-      </ul>
+      <client-only>
+        <ul class="mt-[48px]">
+          <li
+            v-for="({ classId, storyTitle, storyDescription }, index) in nfts"
+            :id="classId"
+            :key="classId"
+            :class="{ 'mt-[88px]': index > 0 }"
+          >
+            <NFTCampaignItem
```
williamchong, line 84: Can we just `client-only` `NFTCampaignItem`?

williamchong (approved):

PR #1086, opened 2023-05-15
✨ Implement NFT book storefront

nwingt:


AuroraHuang22 (approved):

src/components/NFTCover.vue
```diff
@@ -83,24 +98,33 @@ export default {
...
+            'h-full min-h-[290px] max-w-[200px]': this.isNftBook,
           },
         ],
         loading: 'lazy',
         ...this.$attrs,
       };
     },
     resizedSrc() {
-      return getLikeCoResizedImageUrl(this.src, this.size);
+      return getImageResizeAPI(this.src, { width: this.size });
```
williamchong, line 112: `getLikeCoResizedImageUrl` only works for image served by like.co, while `getImageResizeAPI` was for old likerland og
Maybe we should merge both into one function

src/components/Footer.vue
```diff
@@ -38,7 +38,7 @@
           <a class="hover:underline" target="_blank" href="https://like.co">{{ $t('footer_nav_about_likecoin') }}</a>
         </div>
         <div class="grid grid-flow-row gap-y-[16px]">
-          <NuxtLink class="hover:underline" :to="localeLocation({ name: 'writing-nft' })">{{ $t('footer_nav_writing_nft') }}</NuxtLink>
+          <NuxtLink class="hover:underline" :to="localeLocation({ name: 'index' })">{{ $t('footer_nav_writing_nft') }}</NuxtLink>
```
williamchong, line 41: Should this go to `writing-nft-about` instead?

src/pages/index.vue
```diff
@@ -1,7 +1,314 @@
...
+      'pb-[132px]',
+    ]"
+  >
+    <section
+      :class="[
+        'w-full',
+        'max-w-[960px]',
+      ]"
+    >
+      <h2 class="text-[#3AB7A2] text-[48px] font-proxima font-[600]">{{ $t('home_section_book_title') }}</h2>
```
williamchong, line 16: Have book and article in left and right tabs seems better, the current page is too long and articles(WNFT) is too far down below

nwingt: Updated as @joshkiu design

image
image
image

src/components/NFTCover.vue
williamchong: For writing nfts they might, since uri override exists

nwingt: So the criteria for `getLikeCoResizedImageUrl` is URL includes `like.co`?

williamchong: Yes we can do that for now

williamchong (approved):
Please fix lint warning before merge

PR #1089, opened 2023-05-17
💬 Separate tx title for collect & transfer

nwingt:


src/locales/zh-Hant.json
```diff
@@ -567,7 +571,7 @@
   "tx_modal_status_complete_text_collect_portfolio": "你的書架",
   "tx_modal_status_complete_text_collect_without_wallet": "你將收到系統的電郵，請跟隨電郵中的指示領取你的 Writing NFT。",
   "tx_modal_status_complete_text_transfer": "已成功發送你的珍藏",
-  "tx_modal_status_complete_title": "付款成功",
+  "tx_modal_status_complete_title": "交易成功！",
```
williamchong, line 574: I think we changed this string to avoid the word transaction for the whole collect ux

src/locales/en.json
```diff
@@ -577,7 +581,7 @@
   "tx_modal_status_complete_text_collect_portfolio": "your portfolio",
   "tx_modal_status_complete_text_collect_without_wallet": "You will receive an email from us soon. Please follow the instructions in the email to claim your Writing NFT.",
   "tx_modal_status_complete_text_transfer": "Your Writing NFT is successfully transferred!",
-  "tx_modal_status_complete_title": "Payment is successful!",
+  "tx_modal_status_complete_title": "Transaction is complete!",
```
williamchong, line 584: I think we changed this string to avoid the word transaction for the whole collect ux

nwingt: This is the fallback message, collect modal should be using `collect_modal_complete_title`

src/locales/zh-Hant.json
nwingt: This is the fallback message, collect modal should be using `collect_modal_complete_title`

AuroraHuang22 (approved):

williamchong (approved):

PR #1093, opened 2023-05-18
💫 Add poster and transition for pending model viewer

williamchong:
Add poster image
Add transition


src/components/EventModal/Collect/index.vue
```diff
@@ -20,94 +20,96 @@
...
-                  'opacity-0',
-                  'transition-all',
-                  'absolute',
-                  'inset-0',
-                  isFollowPromptStateAuto
-                    ? '!bg-shade-gray text-dark-gray'
-                    : '!bg-like-cyan-light text-like-green',
-                ]"
+    <template #message>
+      <transition name="fade">
```
nwingt, line 24: I think the `<Transition>` element only support one child
should it be wrapping `<model-viewer/>` in this case?

nwingt (approved):

PR #1097, opened 2023-05-19
✨ Adjust NFT Class details page for NFT Book

nwingt:
Screenshot 2023-05-19 at 17 04 32


src/store/modules/nft.js
```diff
@@ -8,9 +8,11 @@ import {
   checkIsWritingNFT,
   isValidHttpUrl,
   formatOwnerInfoFromChain,
+  getISCNRecord,
   getNFTsRespectDualPrefix,
   getNFTClassesRespectDualPrefix,
   getNFTClassCollectionType,
+  getISCNRecord,
```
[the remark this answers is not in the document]
nwingt: Strange that `deploy/rinkeby` missing this import, should be caused by merge conflict

src/store/modules/nft.js
```diff
@@ -134,6 +143,7 @@ const getters = {
   getNFTClassListingInfoById: state => id => state.listingInfoByClassIdMap[id],
   getNFTClassMetadataById: state => id => state.metadataByClassIdMap[id],
   getNFTClassOwnerInfoById: state => id => state.ownerInfoByClassIdMap[id],
+  getNFTClassPriceInfoById: state => id => state.priceInfoByClassIdMap[id],
```
williamchong, line 146: potentially confusing that this function only contains `listing` and `fiat`, but not price info from wnft...

williamchong (approved):

src/store/modules/nft.js
nwingt: Updated

williamchong (approved):

PR #1098, opened 2023-05-19
⚡️ Lazy fetch details for recommendation list

nwingt:
The current recommendation list requires fetching all NFT classes created or collected by the creator, which results in retrieving all corresponding details that lead to an excessive number of API calls. This is also the reason why the creator's information failed to be fetched sometimes.

The fix is to fetch only the necessary recommendation items when they become visible on the screen.


williamchong (approved):
We probably need major refactor for the nft store, we can't rely on fetching everything just to display a class

PR #1103, opened 2023-05-23
⚡️ Remove dual prefix fetch

williamchong:
afaik:
- chain api support dual prefix query natively
- indexer api support dual prefix long time ago, some time after frontend patch was applied
so no reason for frontend to query dual prefix now, we would save 50% api calls here


nwingt (commented):
Please remove debug log

WeiJun0827 (approved):

PR #1106, opened 2023-05-29
✨ Open pdf in browser

williamchong:
image


src/components/NFTPage/PreviewCard.vue
```diff
@@ -284,7 +284,14 @@ export default {
     handleClickViewContent() {
       this.$emit('view-content');
     },
-    handleClickViewContentUrls(type) {
+    handleClickViewContentUrls(e, contentUrl) {
+      const type = this.getContentUrlType(contentUrl);
```
nwingt, line 288: Besides `sameAs` field, do we need to handle if the content URL is a PDF too?

williamchong: UI only shows the "PDF" button if it's in sameAs
Also we dont really know if it's a pdf unless url explicitly end with .pdf
But we can add this for content url too if needed

src/pages/reader.vue
```diff
@@ -0,0 +1,50 @@
+<template>
+  <div class="h-screen">
```
nwingt, line 2: `h-screen`(100vh) behavior is strange in iOS, should hack Tailwind CSS to use 100dvh with 100vh as fallback

https://github.com/likecoin/liker-land/assets/17264716/194ef198-43a0-4c78-ba0a-6412c692b9e8

https://stackoverflow.com/questions/75079019/tailwind-css-fallback-for-new-screen-length-types-such-as-lvh-svh

nwingt (approved):
Tested OK


https://github.com/likecoin/liker-land/assets/17264716/a458a683-3029-45f4-8dba-bc12bb8901e2

PR #1107, opened 2023-05-31
🔧 Use like.co WalletConnect server

williamchong:


nwingt (approved):
Please note that this config also apply to Keplr Mobile and Cosmostation app

PR #1113, opened 2023-06-02
💄 Move notification icon onto site header

williamchong:
For some reason using `append` slot for ButtonV2 here does not work at all
image


nwingt:
Made some adjustments

image

AuroraHuang22 (approved):

PR #1124, opened 2023-06-09
💄 Update UI of collectModal based on new design

AuroraHuang22:
- Logout
截圖 2023-06-09 上午11 47 08

- Login (not sure if we should unify the LIKE price to string ['30% cheaper'](https://www.figma.com/file/xzppAhJVypKBj5vKMn1CT4/%23DePub-app.like.co?type=design&node-id=14547-53245&t=ijT7syKZOwaTewsJ-4).)
截圖 2023-06-09 上午11 56 54

- Login + InsufficientLIKE
截圖 2023-06-09 上午11 57 09


src/components/EventModal/Collect/index.vue
```diff
@@ -12,12 +13,17 @@
...
+        <Label :text="$t('nft_details_page_label_owning')" class="text-like-cyan">
+          <template #prepend>
+            <IconCreativeWork />
+          </template>
+          <template #append>
+            {{ userCollectedCount }}
+          </template>
+        </Label>
+      </div>
     </template>
```
nwingt, line 27: Can we have a slot in TxModal so that we don't have to use absolute position

src/components/EventModal/Collect/MethodButton.vue
```diff
@@ -46,6 +46,11 @@ export default {
       default: false,
     },
   },
+  computed: {
+    getButtonPreset() {
```
nwingt, line 50:
```suggestion
    buttonPreset() {
```

nwingt (approved):

PR #1136, opened 2023-06-19
💄 [_classId/] Add <NFTBookEditionCover/>

AuroraHuang22:
網頁擷取_19-6-2023_22822_rinkeby liker land

https://github.com/likecoin/liker-land/assets/75730405/b6df20d3-0230-4893-aff2-c65606c8c151

### DIFF
--- a/src/components/NFTBook/EditionCover.vue
+++ b/src/components/NFTBook/EditionCover.vue
@@ -1,0 +1,105 @@
+<template>
+  <NFTGemWrapper :is-nft-book="true" :edition-color="themeColor" :class-id="classId">
+    <template v-slot="gem">
+      <NFTPortfolioCard
+        :gem-level="gem.level"
+        :hover-class="gem.hoverClass"
+      >
+        <div class="relative flex items-center justify-center py-[16px]">
+          <div class="absolute inset-0 w-full h-full">
+            <img :src="src" :alt="alt" class="h-auto max-w-full blur-sm">
+          </div>
+          <div
+            class="absolute inset-0 w-full h-full"
+            :style="{
+              backgroundColor: getThemeColor || 'none',
+              opacity: 0.3
+            }"
+          />
+          <div class="relative">
+            <NFTCover
+              :src="src"
+              :size="450"
+              :is-nft-book="true"
+              :alt="alt"
+              :spine-color1="getSpineColor1"
+              :spine-color2="getSpineColor2"
+            />
+          </div>
+        </div>
+        <div
+          :class="[
+            'flex',
+            'flex-col',
+            'text-center',
+            'whitespace-pre-line',
+            'px-[24px]',
+            'pt-[16px]',
+            'pb-[24px]',
+            'bg-white',
+            'relative',
+          ]"
+        >
+          <Label align="center" :text="getNftName" class="text-like-green" />
+          <div class="my-[24px] text-center text-[14px] text-dark-gray">{{ getNftDescription }}</div>
+          <div class="flex items-center justify-center">
+            <ButtonV2 preset="secondary" :text="getPriceLabel" @click="$emit('click-collect',getNftValue)">
+              <template #prepend>
+                <NFTWidgetIconInsertCoin class="w-[16px]" />
+              </template>
+            </ButtonV2>
+          </div>
+        </div>
+      </NFTPortfolioCard>
+    </template>
+  </NFTGemWrapper>
+</template>
+<script>
+export default {
+  props: {
+    src: {
+      type: String,
+      default: '',
+    },
+    classId: {
+      type: String,
+      default: '',
+    },
+    size: {
+      type: Number,
+      default: 720,
+    },
+    alt: {
+      type: String,
+      default: '',
+    },
+    editionConfig: {
+      type: Object,
+      default: () => {},
+    },
+  },
+  computed: {
+    getThemeColor() {
+      return this.editionConfig?.style?.themeColor;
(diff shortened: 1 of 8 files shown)


src/components/NFTBook/EditionCover.vue
```diff
@@ -0,0 +1,105 @@
...
+    },
+    getNftDescription() {
+      return this.editionConfig?.description;
+    },
+    getNftValue() {
+      return this.editionConfig?.value;
+    },
+    getPriceLabel() {
+      return this.editionConfig?.priceLabel;
+    },
```
nwingt, line 102: no need to prefix `get`

src/mixins/nft.js
```diff
@@ -50,6 +50,8 @@ const nftClassCollectionMixin = createNFTClassCollectionMixin({
   typeKey: 'nftClassCollectionType',
 });
 
+const defaultThemeColor = ['', '#FFC123', '#ECBDF3'];
```
nwingt, line 53: empty string means no color? Please add comments

src/pages/nft/class/_classId/index.vue
```diff
@@ -581,11 +602,13 @@ export default {
       );
       return this.handleCollect();
     },
-    handleCollectFromEdition(selectedValue) {
+
+    handleCollectFromEdition(selectedValue, actionType) {
```
nwingt, line 606: I think it is cleaner to separate the methods without adding extra param for `actionType`
```js
handleCollectFromEdition(selectedValue) {
  ...
},

handleCollectFromEditionSelector(selectedValue) {
  this.handleCollectFromEdition(selectedValue);
  logTrackerEvent(....);
},
handleCollectFromEditionTable(selectedValue) {
  this.handleCollectFromEdition(selectedValue);
  logTrackerEvent(....);
}
```

src/components/NFTEditionSelect.vue
```diff
@@ -121,7 +122,7 @@ export default {
       );
     }
     return {
-      selectedValue: selectedItem.value || this.value,
+      selectedValue: selectedItem?.value || this.value,
```
nwingt, line 125: What scenario makes `selectedValue` empty? I think it is a bug to be empty

src/components/NFTBook/EditionCover.vue
```diff
@@ -0,0 +1,105 @@
...
+            'pt-[16px]',
+            'pb-[24px]',
+            'bg-white',
+            'relative',
+          ]"
+        >
+          <Label align="center" :text="getNftName" class="text-like-green" />
+          <div class="my-[24px] text-center text-[14px] text-dark-gray">{{ getNftDescription }}</div>
+          <div class="flex items-center justify-center">
+            <ButtonV2 preset="secondary" :text="getPriceLabel" @click="$emit('click-collect',getNftValue)">
```
nwingt, line 46:
```suggestion
            <ButtonV2 preset="secondary" :text="getPriceLabel" @click="$emit('click-collect', getNftValue)">
```

src/components/NFTBook/EditionCover.vue
```diff
@@ -0,0 +1,105 @@
...
+            </ButtonV2>
+          </div>
+        </div>
+      </NFTPortfolioCard>
+    </template>
+  </NFTGemWrapper>
+</template>
+<script>
+export default {
+  props: {
```
nwingt, line 59: I think we should rename the component name, because it is more than just cover
```suggestion
  name: 'NFTBookEditionCompareTableColumn',
  props: {
```

src/mixins/nft.js
```diff
@@ -294,22 +296,45 @@ export default {
...
+            }
+            const description = edition.description
+              ? edition.description[locale]
+              : '';
+            const priceLabel = formatNumberWithUnit(edition.price, 'USD');
+            const { stock } = edition;
+            const style = {
+              spineColor1: edition.spineColor1 || '#EBEBEB',
+              spineColor2: edition.spineColor2 || '#9B9B9B',
+              themeColor: edition.themeColor || defaultThemeColor[index],
```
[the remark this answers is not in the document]
AuroraHuang22: I think it's unrelated to rarity, just that there can be a clear distinction on the UI.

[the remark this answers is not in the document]
williamchong: Hmm what do we expect to see if there is more than 3 edition?
Also what is the use of `defaultEdition` ?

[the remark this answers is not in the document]
AuroraHuang22: - @nwingt Do you think we should keep defaultThemeColor, or do you have any other suggestions?

- `defaultEdition` is to ensure that listings in the secondary market won't break the UI.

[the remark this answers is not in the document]
williamchong: If we are to keep `defaultThemeColor`, `index % 3` should be used

src/mixins/nft.js
```diff
@@ -147,15 +147,18 @@ export default {
...
+        const { price, nftId, seller } = list[0];
+        return {
+          price,
+          totalPrice: price,
+          classId: this.classId,
+          nftId,
+          seller,
+        };
+      }
+      return undefined;
```
williamchong, line 161: I think the old code is actually bad (it returns `{}` instead of `undefined` if there is no listing)
But returning `undefined` now would be breaking change, so we better return `{}` for now

williamchong (changes requested):

williamchong (approved):

PR #1137, opened 2023-06-19
✨ Support batch collect

nwingt:


src/pages/shopping-cart.vue
```diff
@@ -0,0 +1,263 @@
...
+          <div class="col-span-4 sm:col-span-2 text-like-green font-proxima font-[600] leading-1 text-[2em]">{{ totalNFTPrice | formatNumberWithLIKE }}</div>
+          <template v-if="getAddress">
+            <div class="col-start-2 sm:col-start-4 sm:col-span-1 text-gray-4a">{{ $t('shopping_cart_list_wallet_balance') }}</div>
+            <div
+              :class="[
+                'col-span-4 sm:col-span-2 font-proxima font-[600] leading-1',
+                isInsufficientAmount ? 'text-danger' : 'text-like-green',
+              ]"
+            >{{ walletLIKEBalance | formatNumberWithLIKE }}</div>
+          </template>
```
williamchong, line 79: Very confusing to have wallet balance under total price, should only show if `isInsufficientAmount`

src/components/EventModal/Collect/index.vue
```diff
@@ -564,6 +636,17 @@ export default {
...
+    handleAddToCart() {
+      logTrackerEvent(
+        this,
+        'NFT',
+        'nft_collect_modal_click_add_to_cart',
+        this.classId,
+        1
+      );
+      this.addNFTClassToShoppingCart({ classId: this.classId });
+      this.isAddedToShoppingCart = true;
```
williamchong, line 648: Need to set `this.paymentMethod` before triggering tx, otherwise there will be a close button that allow user to close the dialog too early

src/pages/shopping-cart.vue
```diff
@@ -0,0 +1,263 @@
...
+          });
+          this.uiSetTxError('INSUFFICIENT_BALANCE');
+          return;
+        }
+
+        this.uiToggleCollectModal({
+          classId: this.classIdList,
+          status: TX_STATUS.SIGN,
+        });
+
```
williamchong, line 210: Maybe we should set paymentMethod for collect modal here somehow

src/pages/shopping-cart.vue
```diff
@@ -0,0 +1,263 @@
...
+        this.uiSetTxStatus(TX_STATUS.PROCESSING);
+
+        const { txHash, code } = await broadcastTx(signData, this.getSigner);
+        if (code !== 0) {
+          throw new Error(`TX_FAILED_WITH_CODE_${code}`);
+        }
+        if (!txHash) {
+          throw new Error('TX_FAILED_WITHOUT_TX_HASH');
+        }
+
```
williamchong, line 226: Please set tx status processing non blocking

williamchong (changes requested):
need to handle parsing of `(multiple purchase)` memo to creator message

src/pages/shopping-cart.vue
nwingt: Updated

src/pages/shopping-cart.vue
nwingt: Purchase should be blocking?

williamchong: The part we must block to prevent sequence/grant error is the authz grant tx, after the tx is completed and api received the request with tx hash, the remaining flow is automated

src/components/EventModal/Collect/index.vue
nwingt: Updated

williamchong (approved):
Please update `購物籃` to `背包` & `shopping cart` to `backpack`

williamchong (approved):

PR #1138, opened 2023-06-21
💄 Display view content in dashboard collected tab

williamchong:
image


src/components/NFTPortfolio/Item.vue
```diff
@@ -74,9 +77,12 @@ export default {
     };
   },
   computed: {
-    nftIdForDetails() {
+    isCollectedTab() {
       return this.portfolioTab === 'collected' && this.nftId;
```
nwingt, line 81:
```suggestion
      return this.portfolioTab === 'collected';
```

nwingt (changes requested):

PR #1139, opened 2023-06-21
🐛 Fix wallet selection dialog close on first time

nwingt:
Requires https://github.com/likecoin/likecoin-wallet-connector/pull/26


williamchong (approved):
Please deploy when you see fit

PR #1142, opened 2023-06-23
💄 Update collect modal based on new design

nwingt:
image


nwingt:
@williamchong @AuroraHuang22 Should we just cherrypick part of this commit except for the credit card selection part?

PR #1150, opened 2023-06-27
⬆️ Upgrade @likecoin/wallet-connector to v0.21.0-beta.4

nwingt:


src/components/Dialog.vue
```diff
@@ -152,7 +152,7 @@ export default class Dialog extends Vue {
     return [
       'absolute',
       'inset-x-0',
-      'z-[1000]',
+      'z-[50]',
```
nwingt, line 155: To prevent our dialog cover WalletConnect modal

PR #1151, opened 2023-06-27
🚸  Display the count of nfts reserved or collected by creator

AuroraHuang22:
collect own nft
截圖 2023-06-27 下午9 38 14

collect 0 nft
截圖 2023-06-27 下午9 39 20

reserve nft
截圖 2023-06-27 下午9 42 45

### DIFF
--- a/src/components/NFTPage/CollectorList/index.vue
+++ b/src/components/NFTPage/CollectorList/index.vue
@@ -11,7 +11,7 @@
       <IconPerson />
     </template>
     <template #content="{ contentPaddingClass }">
-      <template v-if="ownerCount">
+      <template v-if="trimmedItems.length">
         <ul :class="contentPaddingClass">
           <NFTPageCollectorListItem
             v-for="owner in trimmedItems"


src/components/NFTPage/CollectorList/index.vue
```diff
@@ -11,7 +11,7 @@
       <IconPerson />
     </template>
     <template #content="{ contentPaddingClass }">
-      <template v-if="ownerCount">
+      <template v-if="trimmedItems.length">
```
williamchong, line 14: So ownerCount does not count owner, but `trimmedItems` actually contains owner's item?
So if there is a WNFT that is owned by both owner + user, owner will be shown in prod now ?

williamchong (approved):

src/components/NFTPage/CollectorList/index.vue
AuroraHuang22: `ownerCount` has already filtered out iscnOwner which is used to display on the user stats.
The source of `trimmedItems` is `getNFTClassOwnerInfoById`, here only the likerland api wallet is filtered out.

> So if there is a WNFT that is owned by both owner + user, owner will be shown in prod now ?

yes， see [https://liker.land/en/nft/class/likenft17ws22xjgsy402x0r59taxp6ap96hmf8meqqcwd3zg2z3c0fzaq0snt4znu](https://liker.land/en/nft/class/likenft17ws22xjgsy402x0r59taxp6ap96hmf8meqqcwd3zg2z3c0fzaq0snt4znu)

PR #1168, opened 2023-06-30
⚡️ Support aggregated proxy API for NFT class info

WeiJun0827:


functions/api/config/config.js
```diff
@@ -14,7 +16,15 @@ config.COOKIE_SECRET = process.env.COOKIE_SECRET;
...
-config.EXTERNAL_URL = process.env.EXTERNAL_URL;
+config.LIKECOIN_CHAIN_API = IS_TESTNET
+  ? 'https://node.testnet.like.co'
+  : 'https://mainnet-node.like.co';
+config.LIKECOIN_API_BASE = IS_TESTNET
+  ? 'https://api.rinkeby.like.co'
+  : 'https://api.like.co';
+config.EXTERNAL_URL = IS_TESTNET
+  ? 'https://rinkeby.liker.land'
+  : 'https://liker.land';
```
williamchong, line 27: Please use env and set `.env` files respectively

src/mixins/nft.js
```diff
@@ -592,9 +593,8 @@ export default {
     async fetchISCNMetadata() {
       await this.lazyGetISCNMetadataById(this.iscnId);
     },
-    async updateNFTClassMetadata() {
-      await catchAxiosError(this.fetchNFTClassMetadata(this.classId));
-      this.lazyGetUserInfoByAddresses(this.iscnOwner);
+    async updateNFTClassAggregatedInfo() {
+      await catchAxiosError(this.fetchNFTClassAggregatedInfo(this.classId));
```
williamchong, line 597: Do we need to call `lazyGetUserInfoByAddresses` at some point? Maybe keep `updateNFTClassMetadata` and call `updateNFTClassAggregatedInfo` and `lazyGetUserInfoByAddresses`?

src/server/config/config.js
```diff
@@ -17,7 +19,15 @@ config.CRISP_USER_HASH_SECRET = process.env.CRISP_USER_HASH_SECRET;
...
-config.EXTERNAL_URL = process.env.EXTERNAL_URL;
+config.LIKECOIN_CHAIN_API = IS_TESTNET
+  ? 'https://node.testnet.like.co'
+  : 'https://mainnet-node.like.co';
+config.LIKECOIN_API_BASE = IS_TESTNET
+  ? 'https://api.rinkeby.like.co'
+  : 'https://api.like.co';
+config.EXTERNAL_URL = IS_TESTNET
+  ? 'https://rinkeby.liker.land'
+  : 'https://liker.land';
```
williamchong, line 30: `EXTERNAL_URL` can be `http://localhost:3000` in dev, so please don't hardcode here

src/util/api/index.js
```diff
@@ -347,36 +339,47 @@ export const postUserV2WalletEmail = ({
...
 };
-export const getUserFollowees = wallet => `/api/v2/users/${wallet}/followees`;
+export const getUserFollowees = wallet =>
+  `${EXTERNAL_HOST}/api/v2/users/${wallet}/followees`;
 export const postFollowCreator = ({ wallet, creator }) =>
-  `/api/v2/users/${wallet}/followers?creator=${creator}`;
+  `${EXTERNAL_HOST}/api/v2/users/${wallet}/followers?creator=${creator}`;
 export const getUserNotificationSettingsUrl = () =>
-  `/api/v2/users/notification`;
+  `${EXTERNAL_HOST}/api/v2/users/notification`;
```
williamchong, line 382: Please use a separated PR for adding `EXTERNAL_HOST` to existing PR

src/store/modules/nft.js
```diff
@@ -399,6 +399,35 @@ const actions = {
...
+        info: purchaseInfo,
+      });
+    }
+    if (classData.iscn_owner) {
+      const promise = dispatch(
+        'lazyGetUserInfoByAddress',
+        classData.iscn_owner
+      );
+      // Need to await if the action fires in during SSR
+      if (!process.client) await promise;
```
williamchong, line 428: Better to let the caller test for `process.client` and await, which page would have issue if we don't await these promises?

src/server/api/routes/nft/metadata.js
```diff
@@ -0,0 +1,165 @@
...
+          }
+          throw err;
+        }),
+    ]);
+
+    const owners = ownerInfoRes.data?.owners || [];
+    const listingInput = listingInfoRes.data?.listing || [];
+    const listings = formatAndFilterListing(listingInput, owners);
+    const purchaseInfo = purchaseInfoRes?.data;
+
```
williamchong, line 50: Please set some short cache header if response is 200, 1s maybe

src/server/api/routes/nft/metadata.js
```diff
@@ -0,0 +1,165 @@
...
+  // in case nft chain metadata was not properly parsed
+  if (!(classData.iscn_owner || classData.account_owner)) {
+    if (iscnRecord) {
+      classData.iscn_owner = iscnRecord.owner;
+      classData.iscn_record_timestamp =
+        iscnRecord?.records?.[0]?.recordTimestamp;
+    } else if (parent.account) {
+      classData.account_owner = parent.account;
+    }
+  }
```
williamchong, line 134: We can always use `iscnRecord` here and ignore `classData.iscn_owner` I suppose

williamchong (changes requested):

src/store/modules/nft.js
WeiJun0827: Was copied from [here](https://github.com/likecoin/liker-land/blob/master/src/store/modules/nft.js#L528-L533).
First saw in [this commit](https://github.com/likecoin/liker-land/pull/857/commits/967e95ce00f4862f8bd8cf7bdcd4482e6e86bb10).

williamchong: @nwingt what does this break if no await?

src/server/api/routes/nft/metadata.js
WeiJun0827: Actually curious about why do we cast this schema inside NFT metadata but not access from ISCN metadata store?
Seems very customized?

src/mixins/nft.js
WeiJun0827: Currently `fetchNFTClassAggregatedInfo` always calls `lazyGetUserInfoByAddresses` here:
https://github.com/likecoin/liker-land/pull/1168/files#diff-bc31414c5136bf8f64828fe6722217f2a2dc6a0e66fc238cce705547403f6dc2R423-R426

Is it preferred not to call `lazyGetUserInfoByAddresses` inside `fetchNFTClassAggregatedInfo` but combine them in `updateNFTClassMetadata`?

src/server/api/routes/nft/metadata.js
williamchong: We didnt have to call iscn record when we first started, so putting iscn owner into api saves a call
But now we always call iscn api, so we should use data in iscn api instead

src/mixins/nft.js
williamchong: Oh just keep it the current way then

src/server/api/routes/nft/metadata.js
WeiJun0827: Like the way but am afraid of introducing breaking change here 🌚
Try not to break sth ...

williamchong: The data from iscn is always correct, so as long as we use same key it is not breaking

src/store/modules/nft.js
nwingt: if not await, creator will sometime be missing in NFT Class details page
image

src/server/api/routes/nft/metadata.js
```diff
@@ -0,0 +1,165 @@
...
+      handleRestfulError(req, res, next, err);
+    }
+  }
+});
+
+async function getNFTClassAndISCNMetadata(classId) {
+  const {
+    data: { class: chainMetadata },
+  } = await axios
+    .get(`${LIKECOIN_CHAIN_API}/cosmos/nft/v1beta1/classes/${classId}`)
```
nwingt, line 71: Should use [getChainNFTClassMetadataEndpoint](https://github.com/nwingt/liker-land/blob/beee54060ac33e3b91d4d96cee1664177e194bd7/src/util/api/index.js#L140)

src/server/api/routes/nft/metadata.js
```diff
@@ -0,0 +1,165 @@
...
+      ownerInfoRes,
+      listingInfoRes,
+      purchaseInfoRes,
+    ] = await Promise.all([
+      getNFTClassAndISCNMetadata(classId),
+      axios.get(
+        `${LIKECOIN_CHAIN_API}/likechain/likenft/v1/owner?class_id=${classId}`
+      ),
+      axios.get(
+        `${LIKECOIN_CHAIN_API}/likechain/likenft/v1/listings/${classId}`
```
nwingt, line 33: should reuse [getChainNFTClassListingEndpoint](https://github.com/nwingt/liker-land/blob/beee54060ac33e3b91d4d96cee1664177e194bd7/src/util/api/index.js#L152)

src/server/api/routes/nft/metadata.js
```diff
@@ -0,0 +1,165 @@
...
+      ),
+      axios
+        .get(`${LIKECOIN_API_BASE}/likernft/purchase?class_id=${classId}`)
```
nwingt, line 36: should reuse [getNFTPurchaseInfo](https://github.com/nwingt/liker-land/blob/beee54060ac33e3b91d4d96cee1664177e194bd7/src/util/api/index.js#L96)

src/server/api/routes/nft/metadata.js
```diff
@@ -0,0 +1,165 @@
...
+      classData = { ...classData, ...apiMetadata };
+    }
+  }
+
+  let iscnRecord;
+  let iscnData;
+  const iscnId = parent?.iscn_id_prefix;
+  if (iscnId) {
+    const { data } = await axios
+      .get(`${LIKECOIN_CHAIN_API}/iscn/records/id?iscn_id=${iscnId}`)
```
nwingt, line 111: Should reuse [getISCNRecord](https://github.com/nwingt/liker-land/blob/beee54060ac33e3b91d4d96cee1664177e194bd7/src/util/api/index.js#L166)

src/server/api/routes/nft/metadata.js
WeiJun0827: Seems we would get a `SyntaxError: Cannot use import statement outside a module` error.
Since we're still using the `common JS` type for server, that way we cannot share functions/configs across the repo...
@williamchong do you have some insight on this issue?

williamchong: Firstly we should not require any file outside of `server` folder from any file inside `server` since it causes error on build
Maybe we have have a file in server for apis for now, and figure out the share part later

src/server/api/routes/nft/metadata.js
williamchong: I meant remove `if (!(classData.iscn_owner || classData.account_owner)) {` and always assign `classData.iscn_owner = iscnRecord.owner;` ...

PR #1194, opened 2023-07-06
⚡️ Load class metadata in chunk

williamchong:


src/server/api/routes/nft/metadata.js
```diff
@@ -19,45 +19,19 @@ router.get('/nft/metadata', async (req, res, next) => {
...
-    res.set('Cache-Control', 'public, max-age=1');
-    res.json({
-      classData,
-      iscnData,
-      ownerInfo,
-      listings,
-      purchaseInfo,
-    });
+    if (Array.isArray(classId)) {
+      const data = await Promise.all(classId.map(id => getNFTClassData(id)));
```
WeiJun0827, line 23: Do we need to guard the size of `classId`?

src/store/modules/nft.js
```diff
@@ -413,6 +414,82 @@ const actions = {
...
+            info: purchaseInfo,
+          });
+        }
+        if (classData.iscn_owner) {
+          const userPromise = dispatch(
+            'lazyGetUserInfoByAddress',
+            classData.iscn_owner
+          );
+          userPromises.push(userPromise);
+        }
```
WeiJun0827, line 454: Maybe we can refactor this part, and use it in single and multiple `classId` cases.
Or let API handle multiple `classId` only, so we don't need to handle the single class case?

src/store/modules/nft.js
```diff
@@ -413,6 +414,82 @@ const actions = {
...
+        const userPromise = dispatch(
+          'lazyGetUserInfoByAddress',
+          classData.iscn_owner
+        );
+        // Need to await if the action fires in during SSR
+        if (!process.client) await userPromise;
+      }
+    }
+  },
+  async fetchNFTClassesAggregatedInfo({ commit, dispatch }, classId) {
```
WeiJun0827, line 492: Do we still keep `fetchNFTClassesAggregatedInfo` here?

williamchong: yes this is extra

WeiJun0827 (approved):

PR #1204, opened 2023-07-08
⚡️ Do not load notification details on login

williamchong:
Notifications events are loaded whenever user is logged in or load likerland with a session
Now this is causing huge performance issue for user with loads of notifcations since all NFT metadata are fetched whenever a likerland page is opened

### DIFF
--- a/src/pages/notifications/index.vue
+++ b/src/pages/notifications/index.vue
@@ -83,6 +83,12 @@
           @click.native="handleClickEvent"
           @click.native.stop
         >
+          <client-only>
+            <lazy-component
+              class="absolute inset-0 pointer-events-none -top-full"
+              @show.once="fetchInfo(event)"
+            />
+          </client-only>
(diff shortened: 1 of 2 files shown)


src/pages/notifications/index.vue
```diff
@@ -83,6 +83,12 @@
           @click.native="handleClickEvent"
           @click.native.stop
         >
+          <client-only>
+            <lazy-component
+              class="absolute inset-0 pointer-events-none -top-full"
```
nwingt, line 88: for using `absolute` position, you need a parent (probably is `<LinkV2/>` in this case) with a `relative` position, otherwise the bound of this `<lazy-component/>` will be unpredictable (I guess it is `<body/>`)

nwingt (changes requested):

nwingt (approved):

PR #1212, opened 2023-07-10
📈 Add click class detail avatar event

williamchong:


src/components/NFTPage/PreviewCard.vue
```diff
@@ -228,6 +229,9 @@ export default {
       }
       this.$emit('view-content-url', type);
     },
+    onClickAvatar(e) {
+      this.$emit('click-avatar', e);
```
AuroraHuang22, line 233:
```suggestion
      this.$emit('click-avatar');
```
Why do we need to pass the event into `handleNFTCardClickAvatar`?

williamchong: For event propagation? i.e. If we don't have better thing to pass, can't go wrong with original event

AuroraHuang22 (approved):

PR #1220, opened 2023-07-14
⚡️ [Portfolio] Sort NFT classes based on basic class info

WeiJun0827:


WeiJun0827:
Found an issue:
For sorting by time, referring `price_updated_at` that comes from `nft` table of indexer will ignore NFTs received by transfer.
Probably should let SQL join with `nft_event` to access the timestamp of `/cosmos.nft.v1beta1.MsgSend`.

src/store/modules/nft.js
```diff
@@ -602,56 +562,35 @@ const actions = {
...
-      }
-    });
-    commit(TYPES.NFT_SET_USER_LAST_COLLECTED_TIMESTAMP_MAP, {
-      address,
-      timestampMap,
+    const nftClassIdDataMap = new Map();
+    collected.forEach(c => nftClassIdDataMap.set(c.id, c));
+    created.forEach(c => nftClassIdDataMap.set(c.id, c));
+    nftClassIdDataMap.forEach((classData, classId) => {
+      dispatch('parseAndStoreNFTClassMetadata', { classId, classData });
```
williamchong, line 585: Should verify if the data format in class api and metadata api is matches, and should normalize them if not.

For instance, we use `iscn_owner` a lot in nft mixin, but in class API the field is returned as `owner`
same for `created_at` vs `iscn_record_timestamp`

We should store these data using consistent field names, so that external callers do not need to check if the data is fetched from class API or metadata API

williamchong (changes requested):
Nicely written but need to handle consistency of data committed into a same store when they are fetched from different sources

src/store/modules/nft.js
```diff
@@ -299,13 +282,13 @@ const getters = {
...
-          Y = getters.getNFTClassPurchaseInfoById(b)?.price;
+          X = getters.getNFTClassPurchaseInfoById(a)?.price || nA.price;
+          Y = getters.getNFTClassPurchaseInfoById(b)?.price || nB.price;
           if (X !== Y) break;
         // eslint-disable-next-line no-fallthrough
         case NFT_CLASS_LIST_SORTING.NFT_OWNED_COUNT:
-          X = getters.getNFTClassOwnerInfoById(a)?.[collector]?.length;
-          Y = getters.getNFTClassOwnerInfoById(b)?.[collector]?.length;
+          X = nA.nfts.length;
+          Y = nB.nfts.length;
```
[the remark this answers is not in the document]
WeiJun0827: `nfts` would be a little piece of owner info of a certain owner for a class.
I suggest we can simplify the data return from indexer, like returning the NFT count only, and fetch complete owner info later.
That way we can have a source of truth.

src/store/modules/nft.js
```diff
@@ -318,35 +301,14 @@ const getters = {
...
-        const aIsWritingNFT = checkIsWritingNFT(
-          getters.getNFTClassMetadataById(a.classId)
-        );
-        const bIsWritingNFT = checkIsWritingNFT(
-          getters.getNFTClassMetadataById(b.classId)
-        );
-        if (aIsWritingNFT && !bIsWritingNFT) return -1;
-        if (!aIsWritingNFT && bIsWritingNFT) return 1;
-        return b.timestamp - a.timestamp;
-      }),
```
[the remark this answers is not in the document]
WeiJun0827: I think the original code here is legacy. Real sorting action has already been handled by `getNFTClassListSorterForCollected` and `getNFTClassIdListSorterForCreated`.
https://github.com/likecoin/liker-land/pull/1220/files#diff-bc31414c5136bf8f64828fe6722217f2a2dc6a0e66fc238cce705547403f6dc2R294-R300

src/store/modules/nft.js
WeiJun0827: Currently, `parseAndStoreNFTClassMetadata` ignore above-mentioned fields and store common fields only.
https://github.com/likecoin/liker-land/blob/master/src/store/modules/nft.js#L497-L511
截圖 2023-07-15 上午10 15 09

Should we further include additional fields?

williamchong: We have a computed field base on `this.NFTClassMetadata?.iscn_owner`
https://github.com/likecoin/liker-land/blob/5d9854faaeeef3a5528c5824f8195013d517bec3/src/mixins/nft.js#L178
Does not loading this field causes UI issue? Preferably we should load and normalize this field into class data via this function

PR #1233, opened 2023-07-18
[TAS-50] ⚡️ Split loading collected and created NFT classes

WeiJun0827:


src/store/modules/nft.js
```diff
@@ -562,6 +569,48 @@ const actions = {
...
+    commit(TYPES.NFT_SET_USER_CREATED_CLASSES_MAP, {
+      address,
+      classes: formatted,
+    });
+
+    classes.forEach(c => {
+      dispatch('parseAndStoreNFTClassMetadata', {
+        classId: c.id,
+        classData: c,
+      });
```
williamchong, line 581: Why not only parse top 5 here?

src/mixins/portfolio.js
```diff
@@ -547,27 +558,13 @@ export const createPortfolioMixin = ({
...
-        this.isLoading = true;
-        await Promise.all(blockingPromises);
-      }
-      this.isLoading = false;
+      this.fetchCollectedNFTClassesByAddress(address).then(() => {
+        this.isLoadingCollectedNFTClasses = false;
+      });
+      this.fetchCreatedNFTClassesByAddress(address).then(() => {
+        this.isLoadingCreatedNFTClasses = false;
+      });
```
williamchong, line 566: Just a question: How complex would this become if we only fetch collect/created when the related tab is shown?

src/mixins/portfolio.js
```diff
@@ -454,8 +456,12 @@ export const createPortfolioMixin = ({
     },
   },
   watch: {
-    nftClassListMap(listMap) {
-      if (!listMap) return;
+    nftClassListOfCollected(list) {
+      if (!list) return;
+      this.$nextTick(this.setupPortfolioGrid);
```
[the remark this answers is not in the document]
WeiJun0827: Not quite sure if there's a better way to handle the mechanism here, maybe @nwingt can suggest a better way?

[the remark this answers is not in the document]
williamchong: I guess something to do with this.currentTab

src/mixins/portfolio.js
WeiJun0827: Can do but haven't looked into how we can trigger fetching from UI. Here we concurrently load both but do not block each other, because I think it's common that users may switch between collected/created tabs. Collected can be preloaded when a user is checking created.

src/store/modules/nft.js
WeiJun0827: Can we only parse top 5 here?

src/mixins/portfolio.js
williamchong: Hitting collected api when viewing created tab affects performance of created tab hugely though? In cases like ckxpress and edmondyu

WeiJun0827: Do you mean performance is affected by fetching the collected list? 
Currently, the UI will load aggregated info and display when the created list is ready, and won't regard whether the collected list is still fetching or is ready.

src/store/modules/nft.js
williamchong: misread, please ignore

src/mixins/portfolio.js
```diff
@@ -80,9 +78,9 @@ export const createPortfolioMixin = ({
     isLoading() {
       switch (this.currentTab) {
         case tabOptions.collected:
-          return this.isLoadingCollectedNFTClasses;
+          return !this.getCollectedNFTClassesByAddress(this.wallet);
         case tabOptions.created:
-          return this.isLoadingCreatedNFTClasses;
+          return !this.getCreatedNFTClassesByAddress(this.wallet);
```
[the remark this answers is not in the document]
WeiJun0827: Then the getter will return an empty array, `isLoading` would be `false`.

src/store/modules/nft.js
```diff
@@ -606,6 +626,26 @@ const actions = {
...
+    } else {
+      await dispatch('fetchCreatedNFTClassesByAddress', address);
+    }
+  },
+  async lazyFetchCollectedNFTClassesByAddress({ state, dispatch }, address) {
+    const classesOrPromise = state.collectedNFTClassesByAddressMap[address];
+    if (classesOrPromise) {
+      if (classesOrPromise instanceof Promise) {
+        await classesOrPromise;
+      }
```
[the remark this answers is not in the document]
WeiJun0827: OK, removed if check

PR #1235, opened 2023-07-19
[TAS-46]⚡️ Add lazy fetch class aggregated info

AuroraHuang22:
**Before**

https://github.com/likecoin/liker-land/assets/75730405/4127144b-048a-46cb-878b-53ecf92d7f54

**After**

https://github.com/likecoin/liker-land/assets/75730405/87d5aa78-a586-4a48-baeb-fb517ae01e82


src/store/modules/nft.js
```diff
@@ -392,6 +392,21 @@ const actions = {
...
+      getters.getISCNMetadataById(metadata?.parent?.iscn_id_prefix),
+      getters.getNFTClassOwnerInfoById(classId),
+      getters.getNFTClassListingInfoById(classId),
+    ];
+
+    if (fieldsToCheck.some(value => !value)) {
+      await dispatch('fetchNFTClassAggregatedInfo', classId);
+    }
+
+    return true;
```
nwingt, line 408: Why returning true here? any use case?

AuroraHuang22: No, it can be removed. 🌚

src/mixins/nft.js
```diff
@@ -611,7 +611,7 @@ export default {
       this.lazyGetUserInfoByAddresses(this.iscnOwner);
     },
     async updateNFTClassAggregatedInfo() {
```
williamchong, line 613: changing the name of this function to `lazyFetchNFTClassAggregatedInfo` might be bette, avoid future confusion about how to actually `update` an already fetched info

src/store/modules/nft.js
```diff
@@ -456,7 +469,12 @@ const actions = {
     }
     return info;
   },
-  parseAndStoreNFTClassMetadata({ commit }, { classId, classData = {} }) {
+  parseAndStoreNFTClassMetadata(
+    { commit, getters },
+    { classId, classData = {} }
+  ) {
+    // Do not overwrite aggregated metadata
+    if (getters.getNFTClassMetadataById(classId)) return;
```
williamchong, line 477: to avoid breaking change, we can
```suggestion
    if (getters.getNFTClassMetadataById(classId)) return getters.getNFTClassMetadataById(classId);
```

williamchong (approved):

PR #1238, opened 2023-07-19
[TAS-45] ⚡️ Paginate notification list

nwingt:
image

https://github.com/likecoin/liker-land/assets/17264716/9f2c29ee-12d1-48bd-af72-22be0f303305


williamchong:
Seems the show more button seems show after there is no more things to show?
At least at some point I click on it and nothing will happen, but the button won't go away

williamchong (approved):

PR #1246, opened 2023-07-21
[TAS-54] 📈 Update NFT class details & portfolio events

nwingt:


src/components/NFTPage/Recommendation.vue
```diff
@@ -101,11 +103,17 @@ export default {
...
   methods: {
+    handleHeaderAvatarClick() {
+      this.$emit('header-avatar-click');
+    },
     handleFollowButtonClick() {
-      this.$emit('on-follow-button-click');
+      this.$emit('follow-button-click');
+    },
+    handleItemClick(classId) {
+      this.$emit('item-click', classId);
```
nwingt, line 113: I don't know why this emit cannot be received by the parent, still investigating...

src/components/NFTPortfolio/MainView.vue
```diff
@@ -556,6 +558,12 @@ export default {
       }
       this.$emit('infinite-scroll', this.portfolioTab);
     }, 2000),
+    handleItemClick(classId) {
+      this.$emit('item-click', classId);
```
nwingt, line 562: I don't know why this emit cannot be received by the parent, still investigating...

src/pages/nft/class/_classId/index.vue
```diff
@@ -229,8 +230,10 @@
           :iscn-owner="iscnOwner"
           :is-followed="isFollowed"
           :recommended-list="recommendedList"
-          @on-follow-button-click="handleFollowButtonClick"
-          @on-nft-item-click="handleRecommendedItemClick"
+          @header-avatar-click="handleRecommendationHeaderAvatarClick"
+          @follow-button-click="handleFollowButtonClick"
+          @item-click="handleRecommendedItemClick"
```
nwingt, line 235: I don't know why `item-click` cannot be received, still investigating...

src/pages/_id/index.vue
```diff
@@ -154,6 +154,8 @@
         @portfolio-reset-filter="handleClearFilter"
         @portfolio-input-filter-change-creator="handleCreatorInputFilterChange"
         @portfolio-input-filter-change-keyword="handleKeywordInputFilterChange"
+        @item-click="handlePortfolioItemClick"
```
nwingt, line 157: I don't know why `item-click` cannot be received, still investigating...

src/components/NFTPage/Recommendation.vue
williamchong: Is th function actually triggered? If not maybe we should manually emit some event in nftPortfolioItem's nuxtLink's click.native

nwingt: Shouldn't trigger event in `<NFTPortfolioItem/>` as we need to emit different events for`<NFTPortfolioPage/>`, `<NFTDetailsPage/>` and `<NFTClassDetailsPage/>`.

I have made a temp fix, we can investigate this later so we can start receiving events asap

PR #1250, opened 2023-07-25
[TAS-144]⚗️ Add A/B test for NFT Book collect button

AuroraHuang22:


src/experiments/nft-book-cta-text.js
```diff
@@ -0,0 +1,7 @@
+export default {
+  name: 'nft-book-cta-text',
+  experimentID: 'jVyrPVSrRguBh7QF_AC8Gg',
+  isEligible: ({ route }) =>
+    !process.server && route.name === 'nft-class-classId',
```
williamchong, line 5:
```suggestion
    !process.server && route.name,
```
I think better not filter by `route.name === 'nft-class-classId'`, this filter runs on the first page of the session, which could be other pages. We want all collect button to run this exp no matter the first page

williamchong (changes requested):

williamchong (approved):

nwingt (approved):

PR #1259, opened 2023-07-27
[TAS-172]💄 Add marked for markdown rendering

AuroraHuang22:
Use [marked](https://github.com/markedjs/marked) package for rendering md edition description.
**Before**
網頁擷取_27-7-2023_215436_rinkeby liker land

**After**
網頁擷取_28-7-2023_05248_localhost


src/assets/css/markdownStyle.css
```diff
@@ -0,0 +1,87 @@
+h1 {
```
nwingt, line 1: I think we should add a parent class in the selector, otherwise all others elements will be affected

williamchong (changes requested):
Do we need to update `functions/package.json` etc too?

PR #1261, opened 2023-07-28
[TAS-185] ✨ Support free WNFT

williamchong:
image
image


nwingt:
Perhaps use `<IconPrice/>` for `<EventModalCollectMethodIcon/>` when the type is free

nwingt (approved):

WeiJun0827:
Post comment: can free WNFT be collected by purchase all?
Will the `totalPrice > 0` when including tx fee?
https://github.com/likecoin/liker-land/blob/master/src/pages/_id/index.vue#L412-L413

PR #1262, opened 2023-07-28
🐛 Fix class filter doesn't includes all creators bug

WeiJun0827:
https://github.com/likecoin/liker-land/assets/33746295/029a750a-bcf5-4ba1-b27b-5758daaa68eb

### DIFF
--- a/src/store/modules/nft.js
+++ b/src/store/modules/nft.js
@@ -494,6 +494,11 @@ const actions = {
       uri,
       parent,
       metadata: classMetadata = {},
+      // HACK: owner field only exists in /likechain/likenft/v1/class endpoint,
+      // not in /cosmos/nft/v1beta1/classes/:classId endpoint.
+      // Should be removed after ISCN data refactor
+      // eslint-disable-next-line camelcase
+      owner: iscn_owner,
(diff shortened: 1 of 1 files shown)


src/store/modules/nft.js
```diff
@@ -494,13 +494,18 @@ const actions = {
...
     const formattedMetadata = {
       name,
       description,
       uri,
       ...classMetadata,
       parent,
+      // HACK: owner field only exists in /likechain/likenft/v1/class endpoint,
+      // not in /cosmos/nft/v1beta1/classes/:classId endpoint.
+      // Should be removed after ISCN data refactor
+      owner,
```
williamchong, line 508: Should we rename this to iscn_owner here to reduce code change?

williamchong (changes requested):

src/store/modules/nft.js
WeiJun0827: Seems linter doesn't allow me to use camel case.

williamchong: add a eslint ignore next line for this case

williamchong (approved):

PR #1264, opened 2023-07-31
[TAS-154] ⚡️ Combine all `/class` API of followees in one

WeiJun0827:


williamchong (changes requested):
Note that this would break when we have too many `iscn_owner` and the url is too long
Can we split the API calls if the followee count is > 50?

WeiJun0827:
> Note that this would break when we have too many `iscn_owner` and the url is too long Can we split the API calls if the followee count is > 50?

I would need to be worried about this. 
The length of one qs set(`&iscn_owner=like1xxx`) is 55 chars, and the max length of an URL is [around 2000](https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers). Is it better to set the batch size at 30?

williamchong:
> > Note that this would break when we have too many `iscn_owner` and the url is too long Can we split the API calls if the followee count is > 50?
> 
> EDIT: Below discussion is for browsers. But I think the max allowed length for servers like Nginx could be set to 4k or 8k, so 50 followees would be fine?
> 
> I would need to be worried about this. The length of one qs set(`&iscn_owner=like1xxx`) is 55 chars, and the max length of an URL is [around 2000](https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers). Is it better to set the batch size at 30?

Lets do 30, we are still firing from frontend for now

williamchong (approved):

PR #1278, opened 2023-08-02
[TAS-115]💄 Hide "View Content" button if external_link is empty

AuroraHuang22:
Currently, if there is no `external_url`, it will display the `iscnURL`. However, the metadata section will also have an iscnUrl btn, so I think it would be more intuitive to hide the "View Content" btn directly if there is no `external_url`.

current situation:
截圖 2023-08-02 下午7 25 52
截圖 2023-08-02 下午7 10 57

### DIFF
--- a/src/mixins/nft.js
+++ b/src/mixins/nft.js
@@ -238,7 +238,10 @@ export default {
       return this.nftMetadata.external_url || this.NFTExternalUrl;
     },
     externalUrl() {
-      return this.iscnUrl || this.nftExternalURL;
+      // The link to the NFT Book info may come from the iscn.json.
+      return this.nftIsNFTBook
+        ? this.iscnUrl || this.nftExternalURL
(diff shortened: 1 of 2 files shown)


williamchong:
Prefer to keep the view content button since it is in a much more visible and intuitive location, the one in metadata is very subtle

src/mixins/nft.js
```diff
@@ -238,7 +238,10 @@ export default {
       return this.nftMetadata.external_url || this.NFTExternalUrl;
     },
     externalUrl() {
-      return this.iscnUrl || this.nftExternalURL;
+      // The link to the NFT Book info may come from the iscn.json.
```
williamchong, line 241:
```suggestion
      // Prioritize iscn url for nft book since book info might be in iscn
```

williamchong (approved):

PR #1279, opened 2023-08-02
🐛 Fix issue caused by undefined records

AuroraHuang22:
occurring in the local environment:
截圖 2023-08-02 下午7 55 27


williamchong (approved):

nwingt:
@AuroraHuang22 Previous solution will show the button even if the content fingerprints of that type is empty/undefined

williamchong (approved):

PR #1285, opened 2023-08-04
[TAS-141] ⚡️ Lazy fetch recommendation data

AuroraHuang22:
**Before**
截圖 2023-08-04 下午2 50 36（2）
截圖 2023-08-04 下午2 51 27

**after**
截圖 2023-08-04 下午3 11 16
截圖 2023-08-04 下午3 10 37

### DIFF
--- a/src/pages/nft/class/_classId/index.vue
+++ b/src/pages/nft/class/_classId/index.vue
@@ -222,6 +222,12 @@
           />
         </section>
         <!-- recommend -->
+        <client-only>
+            <lazy-component
+              class="pointer-events-none"
+              @show.once="handleFetchRecommendInfo"
+            />
+          </client-only>
         <div v-if="isRecommendationLoading" class="flex justify-center items-center my-[24px]">
           <ProgressIndicator />
         </div>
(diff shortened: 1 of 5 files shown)


nwingt:
Can we add a min height for the layout so that it won't collapse when loading causing massive UI shift?

AuroraHuang22:
> Can we add a min height for the layout so that it won't collapse when loading causing massive UI shift?

https://github.com/likecoin/liker-land/assets/75730405/19aefe18-65c2-4eac-92d6-31c1a3deb1be

nwingt (approved):

src/pages/nft/class/_classId/index.vue
```diff
@@ -222,14 +222,17 @@
...
         <!-- recommend -->
-        <div v-if="isRecommendationLoading" class="flex justify-center items-center my-[24px]">
-          <ProgressIndicator />
-        </div>
+        <client-only>
+            <lazy-component
+              class="pointer-events-none"
+              @show.once="handleFetchRecommendInfo"
+            />
+          </client-only>
```
nwingt, line 230:
```suggestion
        <client-only>
          <lazy-component
            class="pointer-events-none"
            @show.once="handleFetchRecommendInfo"
          />
        </client-only>
```

nwingt (changes requested):

PR #1294, opened 2023-08-15
🐛 Fix undefined creator message

AuroraHuang22:
Before
截圖 2023-08-15 下午6 51 28

After
截圖 2023-08-15 下午6 50 53

### DIFF
--- a/src/pages/nft/class/_classId/index.vue
+++ b/src/pages/nft/class/_classId/index.vue
@@ -456,9 +456,7 @@ export default {
       return this.NFTPrice !== undefined && this.NFTPrice > 0;
     },
     creatorMessage() {
-      return this.messageList.find(
-        element => element.event === 'mint_nft' || element.event === 'purchase'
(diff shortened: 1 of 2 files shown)


nwingt (approved):

src/pages/nft/class/_classId/index.vue
```diff
@@ -456,9 +456,7 @@ export default {
       return this.NFTPrice !== undefined && this.NFTPrice > 0;
     },
     creatorMessage() {
-      return this.messageList.find(
-        element => element.event === 'mint_nft' || element.event === 'purchase'
-      );
```
nwingt, line 461: should we keep this as a fallback?

AuroraHuang22: I think it's not necessary. The `messageList` is mainly for showing messages from collectors. Messages from creators should be obtained using `NFTMetadata`.

PR #1297, opened 2023-08-17
[TAS-288] ✨ Add newsletter subscription in footer

nwingt:
image
image


williamchong (approved):
This would have larger impact on loading performance than say, a lazy loaded iframe, though

PR #1310, opened 2023-08-25
[TAS-359] 🚸 Show "open PDF/EPUB" if URL has file extension

nwingt:
Before
PDF URL display as 書訊
image

After
Detected URL with `.pdf`, shows `Open PDF`
image
image

### DIFF
--- a/src/components/NFTViewOptionList.vue
+++ b/src/components/NFTViewOptionList.vue
@@ -113,6 +113,7 @@ export default {
   },
   computed: {
     normalizedContentURLs() {
+      // NOTE: Assuming if only `url` is set, it must contain the actual content rather than the book info
       return this.contentUrls.length ? this.contentUrls : [this.url];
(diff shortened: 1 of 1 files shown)


williamchong (approved):
Please add a comment that we are assuming if no content URL is set and a URL is set in ISCN, that URL must contain content for view/download

src/components/NFTViewOptionList.vue
```diff
@@ -112,11 +112,19 @@ export default {
     },
   },
   computed: {
+    normalizedContentURLs() {
+      return this.contentUrls.length ? this.contentUrls : [this.url];
```
nwingt, line 116:
```suggestion
      // NOTE: Assuming if only `url` is set, it must contain the actual content rather than the book info
      return this.contentUrls.length ? this.contentUrls : [this.url];
```

PR #1312, opened 2023-08-26
[TAS-348] ✨ Display collecting expiry time below collect button

WeiJun0827:
截圖 2023-08-26 下午12 16 41
截圖 2023-08-26 上午11 50 02
截圖 2023-08-26 上午11 49 38
截圖 2023-08-26 上午11 51 53


src/mixins/nft.js
```diff
@@ -271,6 +274,35 @@ export default {
...
+          .replace('in ', '') // en
+          .replace('後', ''); // zh
+        return this.$t('nft_collect_expiry_time_near', { duration });
+      }
+      const date = dateTime.toFormat('yyyy-MM-dd');
+      return this.$t('nft_collect_expiry_time_far', { date });
+    },
+    collectExpiryTimeForNear() {
+      return this.isCollectExpiryTimeNear ? this.collectExpiryTime : '';
+    },
```
williamchong, line 305:
```suggestion
    collectExpiryTimeForDisplay() {
      return this.isCollectExpiryTimeNear ? this.collectExpiryTime : '';
    },
```

src/mixins/nft.js
```diff
@@ -271,6 +274,35 @@ export default {
         ? this.listingInfo.price
         : this.purchaseInfo.price;
     },
+    isCollectExpiryTimeNear() {
```
williamchong, line 277:
```suggestion
    shouldDisplayCollectExpiryTime() {
```

src/mixins/nft.js
```diff
@@ -271,6 +274,35 @@ export default {
...
+
+      const timeLeft = collectExpiryAt - Date.now();
+      if (timeLeft <= 0) return '';
+
+      const dateTime = DateTime.fromMillis(collectExpiryAt);
+      if (this.isCollectExpiryTimeNear) {
+        const duration = dateTime
+          .toRelative({ locale: this.$i18n.locale })
+          .replace('in ', '') // en
+          .replace('後', ''); // zh
```
williamchong, line 297: Just stick with english to extract the number/hour, and format it with `$t` instead? otherwise we need to add one `.replace` line for each locale here, which is unexpected when we are adding new locale

williamchong (changes requested):

src/mixins/nft.js
WeiJun0827: Since `collectExpiryTime` field is also for display, I renamed `collectExpiryTimeForDisplay` to `collectExpiryTimeForHighlight`

src/mixins/nft.js
WeiJun0827: Though luxon is `the evolution of Moment.js`, it does not support our use case, I replaced it with day.js.
https://momentjs.com/docs/#/-project-status/recommendations/
https://github.com/moment/luxon/issues/460

williamchong (approved):

src/locales/zh-Hant.json
```diff
@@ -338,6 +338,8 @@
   "nft_class_footer_raw": "原始數據",
   "nft_class_footer_widget": "NFT 小部件",
   "nft_class_uncollectible": "不能收集",
+  "nft_collect_expiry_time_far": "{date} 截止",
+  "nft_collect_expiry_time_near": "倒數 {duration}",
```
nwingt, line 342:
```suggestion
  "nft_collect_expiry_time_far": "{date}截止",
  "nft_collect_expiry_time_near": "倒數{duration}",

```

nwingt (commented):
I think we should refactor the following code to sync all components behavior
```vue
<div class="flex flex-col gap-[8px] justify-center items-center">
  <ButtonV2
    preset="secondary"
    :text="ctaButtonText"
    :is-disabled="!isCollectable"
    @click="handleClickCTAButton"
  >
    <template #prepend>
      <IconPrice />
    </template>
  </ButtonV2>
  <div 
    v-if="collectExpiryTime" 
    class="flex gap-[4px] justify-center items-center" 
    :class="expiryTimeClass"
  >
    <IconClock />
    <div class="ml-[4px]">{{ collectExpiryTime }}</div>
  </div>
</div>
```

nwingt:
@WeiJun0827 I cannot find related code with 已完結

image

WeiJun0827:
> @WeiJun0827 I cannot find related code with 已完結

I didn't add additional code for expired Classes, since API will return `price = -1` and UI will block with `ended/已完結` message, which is sufficient info message for users (whether it's sold out or expired).
Or do we prefer to distinguish the expired state for users further?

src/locales/zh-Hant.json
WeiJun0827: 截圖 2023-08-29 下午7 47 53
截圖 2023-08-29 下午7 49 19

Above shows the effects after removing spaces.
There is a provided space between `22` and `天`
Should we keep the space?

src/components/CollectButton.vue
```diff
@@ -0,0 +1,85 @@
...
+    },
+    isCollectable: {
+      type: Boolean,
+      default: false,
+    },
+    collectExpiryTime: {
+      type: Number,
+      default: 0,
+    },
+    displayNonHighlightedExpiryTime: {
```
nwingt, line 44: How about
```suggestion
    shouldAlwaysDisplayExpiryTime: {
```

src/components/CollectButton.vue
```diff
@@ -0,0 +1,85 @@
...
+      type: Boolean,
+      default: false,
+    },
+  },
+  computed: {
+    isCollectExpiryTimeNear() {
```
nwingt, line 50:
```suggestion
    isCollectExpiringSoon() {
```

src/components/CollectButton.vue
```diff
@@ -0,0 +1,85 @@
...
+      if (this.isCollectExpiryTimeNear) {
+        const duration = dateTime
+          .locale(normalizeLocaleForDayjs(this.$i18n.locale))
+          .fromNow(true);
+        return this.$t('nft_collect_expiry_time_near', { duration });
+      }
+      const date = dateTime.format('YYYY-MM-DD');
+      return this.$t('nft_collect_expiry_time_far', { date });
+    },
+    expiryTimeClass() {
```
nwingt, line 73:
```suggestion
    expiryTimeTextClass() {
```

src/components/NFTPage/CollectCTA.vue
```diff
@@ -43,17 +43,14 @@
...
-      >
-        <template #prepend>
-          <IconPrice />
-        </template>
-      </ButtonV2>
+      <CollectButton 
+        :button-text="ctaButtonText"
+        :is-collectable="isCollectable"
+        :collect-expiry-time="collectExpiryTime" 
+        :expiry-time-class="expiryTimeClass"
```
nwingt, line 50: Isn't this a computed prop of `<CollectButton/>`?

src/components/NFTCampaign/ItemBase.vue
```diff
@@ -107,16 +107,13 @@
...
-            v-else-if="isCollectable"
-            preset="secondary"
-            @click="handleClickCollect"
-          >
-            <template #prepend>
-              <NFTWidgetIconInsertCoin />
-            </template>
-            {{ $t('nft_widget_button_collect') }}
-          </ButtonV2>
+          <CollectExpiryButton
```
nwingt, line 110: ???
```suggestion
          <CollectButton 
```

src/components/CollectButton.vue
WeiJun0827: Ever thinking of calling the prop with the term `always`, but the passed expiry time will not display in our case.
Is it OK to name it `shouldAlwaysDisplayExpiryTime`?

src/components/NFTPage/CollectCTA.vue
WeiJun0827: Yes, legacy code, should be removed

src/components/CollectButton.vue
nwingt: Then how about
```suggestion
    shouldShowExpiryTimeBeforeExpired: {
```

nwingt (approved):

PR #1323, opened 2023-09-07
[TAS-310] 💄 Update feed item UI for batch send

AuroraHuang22:
截圖 2023-09-07 下午3 48 32

requirment:
- batch sent 的時候顯示following (maximum 8 ) 的 receiver ，在 receiver 後加還給了有多少人
- 如果receiver沒有following，就隨機顯示一個有 Liker ID 的人，refresh 後會改變都可


src/components/SocialFeedItem.vue
```diff
@@ -239,14 +286,29 @@ export default {
...
+    },
     shouldShowNftTitle() {
       return this.type !== EVENT_TYPE.PUBLISH;
     },
     shouldShowItemCard() {
       return this.type === EVENT_TYPE.PUBLISH;
     },
-    shouldShowReceiverInfo() {
-      return this.type !== EVENT_TYPE.PUBLISH;
+    shouldShowSendOrCollectReceiverInfo() {
```
nwingt, line 310:
```suggestion
    shouldShowReceiverInfoForSendOrCollect() {
```

src/components/SocialFeedItem.vue
```diff
@@ -239,14 +286,29 @@ export default {
     nftTitle() {
       return this.nftName || this.classId;
     },
+    nftLink() {
```
nwingt, line 289:
```suggestion
    nftTitleRoute() {
```

src/store/modules/wallet.js
```diff
@@ -625,7 +639,7 @@ const actions = {
 
     commit(WALLET_SET_FEED_EVENT_MEMO, {
       txHash,
-      event: { memo },
+      event: { memo, batchSendList },
```
nwingt, line 642: Should we guard setting `batchSendList` if null

src/store/modules/wallet.js
```diff
@@ -575,10 +578,21 @@ const actions = {
     const txHash = categorizedEvent.tx_hash;
     const event = { ...categorizedEvent };
     let memo;
+    let batchSendList = null;
```
williamchong, line 581: Can we just use `event.tx_hash` to group events in `processEvents` without calling an extra API?

src/components/SocialFeedItem.vue
```diff
@@ -257,6 +319,39 @@ export default {
...
+      }));
+
+      return batchSendListWithInfo;
+    },
+    numberOfAlsoSendTo() {
+      let num = 0;
+      if (this.batchSendList?.length > this.formattedBatchSendList.length) {
+        num = this.batchSendList.length - this.formattedBatchSendList.length;
+      }
+      return this.isBatchSendEvent && num;
```
williamchong, line 353: would be nice if we can cast `this.isBatchSendEvent` to `0` when it is `false`

williamchong (changes requested):

src/store/modules/wallet.js
```diff
@@ -563,11 +566,59 @@ const actions = {
...
+          storedBatchSendEvent.batchSendList.push(receiver);
+        }
+      }
+      if (filteredEvents?.length) {
+        filteredEvents.forEach(event => {
+          const existingEvents = state.followeeEventsMap[event.tx_hash];
+          commit(WALLET_SET_FOLLOWEE_EVENTS, {
+            txHash: event.tx_hash,
+            event: { ...existingEvents, ...event },
+          });
```
williamchong, line 598: We can check for `event.tx_hash` in `batchSendEventsMap` here and commit batch send here, instead of looping again in below block

Also we might want to handle cases where `send` and other type of event are mixed under same `tx_hash`, we should still commit non-send event event if we commited batch send

Maybe we should commit batch send version first, then commit other non-batch or non-send events?

src/store/modules/wallet.js
```diff
@@ -569,56 +569,53 @@ const actions = {
...
+      const isBatchSendEvent =
+        storedBatchSendEvent && storedBatchSendEvent.batchSendList?.length > 1;
+      if (isBatchSendEvent) {
+        commit(WALLET_SET_FOLLOWEE_EVENTS, {
+          txHash,
+          event: {
+            ...event,
+            batchSendList: storedBatchSendEvent.batchSendList,
+          },
+        });
```
williamchong, line 604: Since we are inside the event forEach loop, the batchsend event would be incomplete and commited more than once
Should make the commit part outside of the loop, like in another loop i suppose

williamchong (changes requested):

src/locales/en.json
```diff
@@ -204,6 +204,7 @@
   "FeaturedPage": {
     "headerLabel": "Today"
   },
+  "feed_also_send_to": "also send to {num} people",
```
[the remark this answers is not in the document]
AuroraHuang22: A batch send event is always sent to two or more people, If there's only one recipient, it's a regular send event.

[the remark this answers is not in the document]
AuroraHuang22: oops sorry, u r right, This is not about batchSend.length 🙃

williamchong (approved):

nwingt (approved):

PR #1324, opened 2023-09-08
[TAS-405] 💄 Stylize UI for Mid-autumn Festival

nwingt:
### Collect CTA
image
image

### Collect Dialog
image
image
image
image


src/components/EventModal/Collect/index.vue
```diff
@@ -199,14 +229,25 @@
...
+          class="mx-auto mt-[3rem]"
+          :text="$t('nft_collect_modal_method_free')"
+          :is-disabled="mintedFreeNFT"
+          preset="secondary"
+          @click="handleSelectPaymentMethod('free')"
+        />
+        <ul
+          v-else-if="nftIsFree"
+          class="mt-[16px] flex flex-col gap-[16px] mx-auto max-w-[320px] w-full"
+        >
```
williamchong, line 249: Seems more readable if we use <template v-if="nftIsFree"> to wrap these two, but this would also be fine

williamchong (approved):
Would be nice if we can come up with some software engineering pattern that allow us to avoiding stuffing in lots of div wrapped with `v-if="this.isMidAutumnStyle"` everywhere in the future

PR #1327, opened 2023-09-14
[TAS-374] ✨ Support check out shopping cart by card

WeiJun0827:
截圖 2023-09-15 下午12 10 01

---

截圖 2023-09-15 下午12 04 38

### DIFF
--- a/src/store/modules/nft.js
+++ b/src/store/modules/nft.js
@@ -398,7 +398,7 @@ const actions = {
     const { fiatPrice } = await this.$api.$get(
       api.getStripeFiatPrice({ classId })
     );
-    if (Array.isArray(classId)) {
+    if (!Array.isArray(classId)) {
       commit(TYPES.NFT_SET_NFT_CLASS_FIAT_PRICE_INFO, { classId, fiatPrice });
     }
     return fiatPrice;
--- a/src/store/modules/nft.js
+++ b/src/store/modules/nft.js
@@ -33,7 +33,7 @@ const typeOrder = {
 
 const state = () => ({
   iscnMetadataByIdMap: {},
-  fiatPriceInfoByClassIdMap: {},
+  fiatPriceByClassIdMap: {},
   purchaseInfoByClassIdMap: {},
   listingInfoByClassIdMap: {},
   metadataByClassIdMap: {},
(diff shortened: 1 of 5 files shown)


src/store/modules/nft.js
```diff
@@ -395,17 +390,18 @@ const actions = {
...
     if (!classId) return undefined;
-    const { fiatPrice, listingInfo } = await this.$api.$get(
+    const { fiatPrice } = await this.$api.$get(
       api.getStripeFiatPrice({ classId })
     );
-    const data = { fiatPrice, listingInfo };
-    commit(TYPES.NFT_SET_NFT_CLASS_FIAT_PRICE_INFO, { classId, data });
-    return data;
+    if (Array.isArray(classId)) {
+      commit(TYPES.NFT_SET_NFT_CLASS_FIAT_PRICE_INFO, { classId, fiatPrice });
```
williamchong, line 402: Why check for array here but we doesn't handle it with irritated
Iterators? How about the case if it's not array?

WeiJun0827: It's a bit tricky that the current `fiatPrice` for a single class (not array case) includes a base tx fee, but actually, we should further break it down. Could be in this PR or another PR?

williamchong: Where do we display this price in this PR?

WeiJun0827: I was planning displaying fiat price on the button before redirecting to Stripe page.
That way we should fetch `GET /price` API when mounted.
But that will store a very long key into the store, which is not reusable, and should fetch again if any NFT class in the list is removed.

Do you think we should simply store the LIKE/USD rate and the base tx fee provided by API in store, and calculate the fiat price to display on button? Similar logic for purchasing a single class.

(Or leave it intact in this PR, abd refactor the collect modal and the shopping cart functions in some way soon?

williamchong: Just directly fetch the total fiat price once when we enter shopping cart page, don't calculate using like/usd

WeiJun0827: What about if the user removes some classes in the shopping cart?

williamchong: Refetch from the api

williamchong: I think for this we should keep it as is old behaviour, and ignore array for now
so the guard would actually be `if (!Array.isArray(classId)) {`

WeiJun0827: Yes, a typo

PR #1328, opened 2023-09-14
[TAS-324] ✨ Add suggested follow list

AuroraHuang22:
截圖 2023-09-15 上午12 36 42

https://github.com/likecoin/liker-land/assets/75730405/1b10269e-1f4a-4929-9789-b43c8fca395f


src/server/api/routes/users/v2/follow.js
```diff
@@ -14,7 +14,7 @@ const router = Router();
 
 router.get('/followees', authenticateV2Login, async (req, res, next) => {
   try {
-    const { user } = req.session;
+    const user = req.query?.user || req.session.user;
```
williamchong, line 17: Should not replace `req.session.user` in any api that use `authenticateV2Login`, otherwise anyone can access other's information (in this case, followees and pastFollowees) as long as they are logged in.

Should make a new API for suggestion that does not leak followee information of other user

src/store/modules/wallet.js
```diff
@@ -672,6 +680,46 @@ const actions = {
...
+      );
+
+      const followSuggestions = [...followees, ...pastFollowees].filter(
+        f => !currentFolloweesSet.has(f)
+      );
+
+      followSuggestions.forEach(suggestion => suggestedSet.add(suggestion));
+
+      currentIndex += 1;
+    }
```
williamchong, line 716: Should handle in server side since other user's followee information should not be public

williamchong (changes requested):

src/server/api/routes/users/v2/follow.js
```diff
@@ -9,41 +9,61 @@ const {
...
+      let currentIndex = 0;
+
+      // If we don't have enough suggestions, fetch more
+      while (suggestedSet.size < 3 && currentIndex < currentFollowees.length) {
+        // eslint-disable-next-line no-await-in-loop
+        const { followees = [], pastFollowees = [] } = await getUserFollowees(
+          currentFollowees[currentIndex]
+        );
+
+        const followSuggestions = [...followees, ...pastFollowees].filter(
```
williamchong, line 50: Should not suggest `pastFollowees` to users

src/server/api/routes/users/v2/follow.js
```diff
@@ -9,41 +9,61 @@ const {
...
+          f => !currentFolloweesSet.has(f)
+        );
+
+        followSuggestions.forEach(suggestion => suggestedSet.add(suggestion));
+        currentIndex += 1;
```
williamchong, line 55: We would want avoid just leaking the whole follower list of `currentIndex` to user. Thus we should limit and perform some randomization of the output

- Size of `currentFollowees` must be larger than a number, say 3, otherwise following 1 user would always yield result from that particular list
- `currentFollowees` should contain some followees that are not already in `currentFolloweesSet` in order to be counted towards to the size limit above
- We should randomly draw from `currentFollowees` instead of using `currentIndex`

If the above conditions are not met, we should just return default suggestions/nothing.

Also note these rules are still not perfect, a better idea would be to use public information for suggestions, instead of non-public information like follower list i.e. use creators of NFT owned by a user

src/store/modules/wallet.js
```diff
@@ -672,6 +681,34 @@ const actions = {
...
   },
 
+  async fetchSuggestedFollowList({ state, commit }) {
+    const currentFollowees = state.followees;
+    const currentFolloweesSet = new Set(currentFollowees);
+
+    // Cache check and return cached result if it's valid
+    if (
+      state.suggestedFollowList.length &&
+      !currentFollowees.some(user => state.suggestedFollowList.includes(user))
```
williamchong, line 691: Should we remove entry from `suggestedFollowList` if user click follow in suggested list?

PR #1339, opened 2023-09-26
[TAS-458] ✨ Add more NFT Books page

nwingt:
image

localhost_3000_zh-Hant_store_books (1)

image


src/mixins/nft.js
```diff
@@ -184,6 +184,9 @@ export default {
         this.NFTClassMetadata.account_owner
       );
     },
+    iscnAuthor() {
```
williamchong, line 187: this name might cause potential confusion in the future but i have no good suggestion...
`iscnMetadataAuthor` ?

src/pages/store/books.vue
```diff
@@ -0,0 +1,64 @@
...
+        href: `/api/nft/metadata?class_id=${classId}`,
+      });
+    });
+    return {
+      link,
+    };
+  },
+  computed: {
+    nftBooks() {
+      return LIKECOIN_NFT_BOOK_ITEMS.map(classId => ({ classId }));
```
williamchong, line 52: Should this `books` page include featured book too?

src/components/NFTBook/ItemCard.vue
```diff
@@ -44,7 +44,7 @@
     <Label
       class="text-medium-gray mt-[6px] mb-[12px]"
       preset="p6"
-      :text="creatorDisplayName | ellipsis"
+      :text="(iscnAuthor || creatorDisplayName) | ellipsis"
```
AuroraHuang22, line 47: I think it can be a computed property

src/mixins/nft.js
nwingt: How about iscnWorkAuthor that imply Author under CreativeWork in schema.org

src/components/NFTBook/ItemCard.vue
nwingt: can change to computed property if we used it more than once

src/mixins/nft.js
williamchong: makes sense

src/pages/store/books.vue
nwingt: Yes, included

src/mixins/nft.js

williamchong (approved):

PR #1355, opened 2023-10-06
💩 Fix mysterious site menu z-index issue

nwingt:
Maybe related to Nuxt 🤷‍♂️
image


williamchong (approved):
How does this work...

PR #1357, opened 2023-10-09
[TAS-512] 🔍️ Add noindex tag for hidden NFT

williamchong:


WeiJun0827 (approved):

AuroraHuang22 (approved):

src/pages/nft/class/_classId/_nftId.vue
```diff
@@ -523,6 +531,7 @@ export default {
   async mounted() {
     try {
       this.lazyGetUserInfoByAddresses(this.iscnOwner);
+      this.fetchIscnOwnerNFTDisplayStateList();
```
nwingt, line 534: should fetch in server side?

williamchong: Don't want to add another blocking call in fetch just for this feature
since Google actually renders javascript, I think having the `noindex` added to meta on mount might already work

PR #1377, opened 2023-10-25
🔍️ Set i18n in sitemap

williamchong:
Hack mentioned in https://github.com/nuxt-community/sitemap-module/issues/122#issuecomment-633424117

### DIFF
--- a/src/config/sitemap.js
+++ b/src/config/sitemap.js
@@ -10,6 +10,7 @@ const axios = Axios.create({
   maxContentLength: 50 * 1000 * 1000,
 });
 
+const locales = ['zh-Hant', 'en'];
 /* Copied from constant due to nuxt.config.js is not es6 import syntax */
 const { IS_TESTNET } = process.env;
 const LIKE_CO_API = IS_TESTNET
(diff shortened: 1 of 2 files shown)


src/config/sitemap.js
```diff
@@ -65,7 +66,10 @@ async function getSitemapRoutes() {
   const portfolioPageRoutes = [...new Set(creator.concat(collectors))].map(
     id => `/${id}`
   );
-  return nftDetailsPageRoutes.concat(portfolioPageRoutes);
+  return nftDetailsPageRoutes.concat(portfolioPageRoutes).map(url => ({
+    url: `/${locales[0]}${url}`,
```
nwingt, line 70: ???
```suggestion
    url: `/${locales[0]}/${url}`,
```

src/config/sitemap.js
```diff
@@ -65,7 +66,10 @@ async function getSitemapRoutes() {
...
+    links: locales.map(lang => ({ lang, url: `/${lang}${url}` })),
```
nwingt, line 71:
```suggestion
    links: locales.map(lang => ({ lang, url: `/${lang}/${url}` })),
```

src/config/sitemap.js
williamchong: Our `url` starts with / as shown in e.g. line 66

nwingt (approved):

PR #1379, opened 2023-10-30
[TAS-487] 🚸 Allow clicking collect button after sold out

nwingt:
### Before
image
image

### After
image
image

### Remain disabled on details page
image


src/mixins/nft.js
```diff
@@ -774,6 +774,10 @@ export default {
       await this.fetchNFTDisplayStateListByAddress(this.getAddress);
     },
     async collectNFT() {
+      if (!this.nftIsCollectable) {
+        this.$router.push({ query: { ...this.$route.query } });
```
williamchong, line 778: Wouldn't this cause duplicated navigation? potentially breaks the page

williamchong (changes requested):

src/mixins/nft.js

williamchong (approved):

PR #1380, opened 2023-10-30
[TAS-629] 📈 Log user login method to user db

williamchong:


src/store/modules/wallet.js
```diff
@@ -889,9 +889,12 @@ const actions = {
     const { address } = state;
     try {
       commit(WALLET_SET_IS_LOGGING_IN, true);
-      const { signer } = state;
+      const { signer, methodType } = state;
       const data = await signLoginMessage(signer, address);
-      await this.$api.post(postUserV2Login(), data);
+      await this.$api.post(postUserV2Login(), {
+        loginMethod: methodType,
```
nwingt, line 895: is `loginMethod` or `method`?

williamchong: lets use loginMethod

nwingt (approved):

PR #1384, opened 2023-11-01
[TAS-646] 💄 Allow directly claim free book via wallet

williamchong:
Requires https://github.com/likecoin/likecoin-api-public/pull/543

### DIFF
--- a/src/pages/nft/claim/index.vue
+++ b/src/pages/nft/claim/index.vue
@@ -30,39 +30,6 @@
         :is-content-viewable="true"
         @view-content-url="handleClickViewContentDirectly"
       />
-      <template v-if="isFreePurchase">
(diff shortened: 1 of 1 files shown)


src/pages/nft/claim/index.vue
```diff
@@ -30,39 +30,6 @@
...
-        <template v-else-if="state === 'CLAIMED'">
-          <Label
-            class="text-medium-gray"
-            preset="p6"
-            :text="$t('nft_free_claim_check_email_label')"
-            align="center"
-          />
-        </template>
-      </template>
       <template v-else-if="!claimingAddress">
```
nwingt, line 33: Should be this cause of the missing claim button
```suggestion
      <template v-if="!claimingAddress">
```

PR #1386, opened 2023-11-02
[TAS-541] 🚸 Allow copy wallet in feed page

nwingt:
image

### Mobile
Screenshot 2023-11-03 at 02 43 36

### DIFF
--- a/src/components/UserInfoCard.vue
+++ b/src/components/UserInfoCard.vue
@@ -1,0 +1,87 @@
+<template>
+  <CardV2 class="flex flex-col items-center w-full gap-[1rem]">
+    <Identity
+      :class="{ 'cursor-pointer hover:scale-105 transition-transform': $listeners['click-avatar'] }"
+      :avatar-url="avatarSrc"
+      :avatar-size="88"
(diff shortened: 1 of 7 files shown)


src/components/UserInfoCard.vue
```diff
@@ -0,0 +1,87 @@
+<template>
+  <CardV2 class="flex flex-col items-center w-full gap-[1rem]">
+    <Identity
+      :class="{ 'cursor-pointer hover:scale-105 transition-transform': $listeners['click-avatar'] }"
+      :avatar-url="avatarSrc"
```
williamchong, line 5: should also set displayname here?

williamchong, line 5: should also set displayname here?

nwingt: You mean `alt`?

williamchong: Yes it is for `alt` but I think the prop is called `displayName`

nwingt: Updated

williamchong (approved):

PR #1389, opened 2023-11-02
[TAS-543] 🚸  Possibly show claimed for free NFT

nwingt:


williamchong (approved):
One potential refactor is can we put `isFree && !isCollected` judgement in `isCollectable` though

PR #1398, opened 2023-11-14
[TAS-595] 💄 Display nft book buyer messages on collector list

AuroraHuang22:
截圖 2023-11-15 下午2 46 31

### DIFF
--- a/src/mixins/nft.js
+++ b/src/mixins/nft.js
@@ -698,86 +698,86 @@ export default {
         actionType.push('mint_nft');
       }
 
-      try {
-        const ignoreToList = this.nftIsWritingNFT
-          ? LIKECOIN_NFT_API_WALLET
-          : '';
-        let dbEventMap = null;
-        if (this.nftIsWritingNFT) {
-          dbEventMap = await getNFTHistoryDataMap({
-            axios: this.$api,
-            classId: this.classId,
-            nftId: this.nftId,
-          });
-        }
-        const {
-          nextKey,
-          events: latestBatchEvents,
-        } = await getFormattedNFTEvents({
+      const ignoreToList = this.nftIsWritingNFT ? LIKECOIN_NFT_API_WALLET : '';
+      let dbEventMap = null;
+      if (this.nftIsWritingNFT) {
+        dbEventMap = await getNFTHistoryDataMap({
           axios: this.$api,
           classId: this.classId,
           nftId: this.nftId,
-          actionType,
-          ignoreToList,
         });
+      }
+      const {
+        nextKey,
+        events: latestBatchEvents,
+      } = await getFormattedNFTEvents({
+        axios: this.$api,
+        classId: this.classId,
+        nftId: this.nftId,
+        actionType,
+        ignoreToList,
+      });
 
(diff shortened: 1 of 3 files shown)


src/mixins/nft.js
```diff
@@ -715,9 +716,34 @@ export default {
...
       });
-      this.NFTHistory = this.nftIsWritingNFT
-        ? populateGrantEvent(latestBatchEvents, dbEventMap)
-        : latestBatchEvents;
+
+      const nftBookLatestBatchEvents = latestBatchEvents;
+      if (this.nftIsNFTBook) {
+        const { messages: nftBookBuyerMessages } = await this.$api.$get(
+          getNftBookBuyerMessage(this.classId)
+        );
```
williamchong, line 724: Should try catch for 404 in case the book is not listed at all

src/mixins/nft.js
```diff
@@ -715,9 +716,34 @@ export default {
...
+        for (let i = 0; i < nftBookBuyerMessages.length; i += 1) {
+          const buyerMessage = nftBookBuyerMessages[i].message;
+          const { txHash } = nftBookBuyerMessages[i];
```
williamchong, line 727: Note that author can leave a message in this txHash as memo too

src/mixins/nft.js
```diff
@@ -715,9 +716,34 @@ export default {
...
+          }
+        }
+      }
+
+      if (this.nftIsWritingNFT) {
+        this.NFTHistory = populateGrantEvent(latestBatchEvents, dbEventMap);
+      } else if (this.nftIsNFTBook) {
+        this.NFTHistory = nftBookLatestBatchEvents;
+      } else {
+        this.NFTHistory = latestBatchEvents;
```
williamchong, line 744: I think `nftBookLatestBatchEvents` and `latestBatchEvents` are same since array is pass by ref, no need `if else`

src/pages/nft/class/_classId/index.vue
```diff
@@ -573,6 +573,35 @@ export default {
...
       return collectorsWithMemo;
     },
+    populatedBuyerWithMessage() {
+      if (!this.populatedDisplayEvents) {
+        return this.populatedCollectors;
+      }
+      const collectorsWithBuyerMessages = this.populatedCollectors.map(
+        buyer => {
+          const event = this.populatedDisplayEvents.find(
+            event => event.buyerMessage && buyer.id === event.toWallet
```
williamchong, line 583: I think we can make a filtered list `this.populatedDisplayEvents.filter(e => e.buyerMessage)` before the `.map` function and use that instead, for cleaner code and performance

williamchong (changes requested):

williamchong (approved):

src/mixins/nft.js
```diff
@@ -715,9 +717,39 @@ export default {
...
+
+              if (matchingEvent) {
+                matchingEvent.buyerMessage = buyerMessage;
+              }
+            }
+          }
+        }
+      } catch (error) {
+        // eslint-disable-next-line no-console
+        console.error(error);
```
williamchong, line 744: Even better if we check and skip logging for 404

PR #1400, opened 2023-11-15
[TAS-709] 🚸 Download content using sameAs name for filename

nwingt:


src/components/NFTViewOptionList.vue
```diff
@@ -156,9 +158,25 @@ export default {
...
+    async handleClickViewContentURL(e, contentUrl) {
       const type = this.getContentUrlType(contentUrl);
-      this.$emit('view-content-url', e, parseNFTMetadataURL(contentUrl), type);
+      const url = parseNFTMetadataURL(contentUrl);
+      this.$emit('view-content-url', e, url, type);
+      if (type !== 'pdf') {
+        try {
+          this.alertPromptSuccess(this.$t('nft_download_content_prepare'));
+          const res = await fetch(url, { mode: 'no-cors' });
+          const blob = await res.blob();
```
williamchong, line 169: Why not use axios blob mode?

williamchong (changes requested):

src/components/NFTViewOptionList.vue
nwingt: Updated

src/components/NFTViewOptionList.vue
```diff
@@ -156,9 +158,24 @@ export default {
     handleClickViewContent() {
       this.$emit('view-content');
     },
-    handleClickViewContentURL(e, contentUrl) {
+    async handleClickViewContentURL(e, contentUrl) {
       const type = this.getContentUrlType(contentUrl);
-      this.$emit('view-content-url', e, parseNFTMetadataURL(contentUrl), type);
+      const url = parseNFTMetadataURL(contentUrl);
+      this.$emit('view-content-url', e, url, type);
+      if (type !== 'pdf') {
```
williamchong, line 165: Should pdf's `router.push` to reader page be refactored into here also? otherwise actions are all spreaded out into different components
Also consider keeping the `<a>`  with `href` and `download` and use `e.preventDefault()` for more link context and accessibility

nwingt: Updated

williamchong (approved):

PR #1405, opened 2023-11-17
[TAS-528] ✨ Support purchasing NFT Book by LIKE

WeiJun0827:


src/pages/nft/class/_classId/index.vue
```diff
@@ -905,7 +905,13 @@ export default {
         1
       );
     },
-    handleEditionSelectChange() {
+    handleEditionSelectChange(selectedValue) {
+      this.$router.push({
```
williamchong, line 909: Should use .replace()

WeiJun0827: Noted, thanks!!

src/store/modules/nft.js
```diff
@@ -55,9 +55,22 @@ const mutations = {
...
-      Vue.set(state.paymentPriceByClassIdMap, classId, info);
+      if (Number.isInteger(priceIndex)) {
+        if (state.paymentPriceByClassIdMap[classId]) {
+          Vue.set(state.paymentPriceByClassIdMap[classId], priceIndex, info);
+        } else {
+          Vue.set(state.paymentPriceByClassIdMap, classId, {
+            [priceIndex]: info,
+          });
+        }
+      } else {
```
williamchong, line 71: Would just setting the key as `classId_priceIndex` be simpler in structure?

src/pages/nft/class/_classId/index.vue
```diff
@@ -888,14 +883,42 @@ export default {
...
+                free: true,
+                price_index: edition.index,
+                from: 'liker_land_waived',
+              },
+            })
+          );
+          return;
+        }
+        if (edition.price > 0) {
+          await this.fetchNFTPrices();
```
williamchong, line 902: Should fetch when edition is changed instead of onClick, adding too much delay in user action might cause issue in popup/redirect

src/pages/nft/class/_classId/index.vue
```diff
@@ -888,14 +883,42 @@ export default {
...
+          if (this.nftPriceInLIKE > 0) {
+            const gaClientId = await getGaClientId(this);
+            const link = getNFTBookPurchaseLink({
+              classId: this.classId,
+              priceIndex: edition.index,
+              platform: this.platform,
+              gaClientId,
```
williamchong, line 909: Why pop up stripe when `nftPriceInLIKE > 0`?

williamchong (changes requested):

src/pages/nft/class/_classId/index.vue
WeiJun0827: Bad logic...

src/components/EventModal/Collect/index.vue
```diff
@@ -456,6 +503,7 @@ export default {
       justCollectedNFTId: undefined,
       shouldShowMessageInput: false,
       memo: '',
+      email: this.$store.getters.walletEmail || '',
```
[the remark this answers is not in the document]
WeiJun0827: How do we use `this.getAddress`?
I tried use `this.walletEmail` but not work.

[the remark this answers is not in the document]
williamchong: I think `this.walletEmail` should work..
By not working your mean `this.$store.getters.walletEmail` has value but `this.walletEmail` doesn't?

src/components/EventModal/Collect/index.vue
```diff
@@ -747,8 +776,9 @@ export default {
             );
             const result = await this.collectNFTWithLIKE(classId, {
               memo: this.memo,
+              email: this.email,
             });
-            if (result) {
+            if (result && !this.nftIsNFTBook) {
```
williamchong, line 781: Just check for `result.nftId || result.purchased?.[0]?.nftId` is undefined instead of hardcoding book?

src/store/modules/nft.js
```diff
@@ -785,6 +793,34 @@ const actions = {
...
+    { getters, dispatch },
+    { classId, priceIndex }
+  ) {
+    const info = getters.getNFTClassPaymentPriceById(classId, priceIndex);
+    if (!info) {
+      await dispatch('fetchNFTBookPaymentPriceInfoByClassIdAndPriceIndex', {
+        classId,
+        priceIndex,
+      });
+    }
```
williamchong, line 822: Returning `info` for these function might make them more useful in more scenario

src/pages/nft/class/_classId/index.vue
```diff
@@ -888,6 +884,35 @@ export default {
...
+              this,
+              'add_shipping_info',
+              purchaseEventParams
+            );
+            this.fetchUserCollectedCount();
+            this.walletFetchLIKEBalance();
+          }
+          this.uiToggleCollectModal({ classId: this.classId });
+          return;
+        }
```
williamchong, line 915: use `else if` instead of return here might be easier to maintain, since a long block of early returns is harder to read

src/mixins/nft.js
```diff
@@ -1037,6 +1083,16 @@ export default {
...
+      if (this.nftIsNFTBook) {
+        const link = getNFTBookPurchaseLink({
+          classId: this.classId,
+          priceIndex: this.editionPriceIndex,
+          platform: this.platform,
+        });
+        window.open(link, '_blank', 'noopener');
+        return;
+      }
+
```
williamchong, line 1095: use `if else` here?

williamchong (changes requested):

src/components/EventModal/Collect/index.vue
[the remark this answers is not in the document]
WeiJun0827: Yes, `this.$store.getters.walletEmail` works but `this.walletEmail` doesn't

src/store/modules/nft.js
```diff
@@ -785,6 +793,36 @@ const actions = {
...
+    return info;
```
williamchong, line 824:
```suggestion
    let info = getters.getNFTClassPaymentPriceById(classId, priceIndex);
    if (!info) {
      info = await dispatch('fetchNFTBookPaymentPriceInfoByClassIdAndPriceIndex', {
        classId,
        priceIndex,
      });
    }
    return info;
```

PR #1411, opened 2023-11-21
[TAS-787] ✨ Support hideDownload flag in nft book

williamchong:
https://github.com/likecoin/likecoin-api-public/pull/556


nwingt (commented):
can we pick `canDownloadContent` or `isContentDownloadable`?

williamchong:
> can we pick `canDownloadContent` or `isContentDownloadable`?

`isContentDownloadable` then, but do you want to change `isDownloadHidden` too?

nwingt (approved):

PR #1441, opened 2023-12-01
[TAS-825] 💬 Update claim page copywriting

nwingt:
image
image
image


src/locales/en.json
```diff
@@ -370,6 +370,8 @@
   "nft_collect_modal_alert_fail": "Collect {name} fail: {error}",
   "nft_collect_modal_alert_success": "{name} was collected successfully",
   "nft_collect_modal_already_minted": "You have already collected this work.",
+  "nft_collect_modal_email": "Email",
+  "nft_collect_modal_email_placeholder": "Leave email to pay by LIKE",
```
williamchong, line 374: Where are these used?

williamchong (changes requested):

src/locales/en.json
nwingt: Updated, JTM cached

williamchong (approved):

PR #1460, opened 2023-12-13
[TAS-763] Implement PG page UI

AuroraHuang22:


src/plugins/plyr.js
```diff
@@ -0,0 +1,7 @@
+import Vue from 'vue';
+import VuePlyr from 'vue-plyr';
+import 'vue-plyr/dist/vue-plyr.css';
+
+Vue.use(VuePlyr, {
+  plyr: {},
+});
```
williamchong, line 7: Can we load `VuePlyr` only in the gutenberg page as a component, instead of loading this site-wise as a plugin?

PR #1464, opened 2023-12-16
[TAS-886] 🚑 Fix broken iframe due to CSP

AuroraHuang22:


src/nuxt.config.js
```diff
@@ -249,6 +251,7 @@ const nuxtConfig = {
           'https://likecoin.github.io',
           'verify.walletconnect.com',
           'www.facebook.com',
+          '*',
```
williamchong, line 254: Prefer just allowing youtube domain, * basically mean allow all which the the iframe-src list useless

src/nuxt.config.js
```diff
@@ -238,6 +238,8 @@ const nuxtConfig = {
           'https://substackapi.com',
           '*.crisp.chat',
           'connect.facebook.net',
+          'https://www.youtube.com/iframe_api',
+          'https://www.youtube.com/s/player/d23221b6/www-widgetapi.vflset/www-widgetapi.js',
```
williamchong, line 242: Why do we need these script?

williamchong (changes requested):

src/nuxt.config.js
AuroraHuang22: [updated](https://github.com/likecoin/liker-land/pull/1464/commits/3e914f84134f07b720f26cb0272156b045c34b4b)

src/nuxt.config.js
AuroraHuang22: Because the embed calls these APIs, and one of the reasons for the broken display is also due to this.

截圖 2023-12-17 上午10 29 19
截圖 2023-12-17 上午10 29 37

src/pages/gutenberg/index.vue
```diff
@@ -412,6 +420,15 @@ export default {
     handleSliderMove() {
       this.$emit('slider-move');
     },
+    scrollToElement(hash) {
+      const currentRef = this.$refs[hash];
+      setTimeout(() => {
+        currentRef.scrollIntoView({
```
nwingt, line 426:
```suggestion
      setTimeout(() => {
        const currentRef = this.$refs[hash];
        currentRef.scrollIntoView({
```

williamchong (approved):

PR #1465, opened 2023-12-17
[TAS-886]💄 Update [/gutenberg/free-audio-book] page UI

AuroraHuang22:
截圖 2023-12-17 下午4 32 19
截圖 2023-12-17 下午4 19 48


src/locales/zh-Hant.json
```diff
@@ -261,6 +261,13 @@
       "Title": "@:gutenbergPage.Title"
     }
   },
+  "gutenbergFreeAudioBooksPage":{
+    "Title": "下載免費有聲書",
+    "Og": {
+      "Description": "古騰堡計劃、微軟和麻省理工學院的合作，將 5,000 本公有領域的著作製作成有聲書，並由 LikeCoin 去中心化社群保存。這個區塊鏈出版革命讓公有知識更易普及，同時促進知識的共享和保護。探索分散式出版的新時代。",
+      "Title": "@:gutenbergFreeAudioBooksPage.Title"
+    }
+  },
```
nwingt, line 270: We prefer not to use nested key, which is difficult to find and replace

all keys should be in `snake_case`

nwingt (approved):

PR #1480, opened 2023-12-29
[TAS-871] 🎨 Add ellipsis for iscnWorkAuthor

AuroraHuang22:
Before
截圖 2023-12-29 下午2 28 29

After
截圖 2023-12-29 下午2 02 45


nwingt (commented):
`truncate` should work, maybe due to `flex` parent, try to apply truncate to its parent and its grandparent

AuroraHuang22:
> `truncate` should work, maybe due to `flex` parent, try to apply truncate to its parent and its grandparent
截圖 2023-12-29 下午2 51 53

I just found that we can easily resolve this issue by removing the `truncate `

nwingt (approved):

PR #1485, opened 2024-01-05
🐛 Fix incorrect chapter path

WeiJun0827:
Note: The TOC list utilizes the HTML or XHTML filenames of chapters as its keys. 
However, to navigate to a specific chapter, the `rendition.display` function requires the full path of the file. 
It's important to acknowledge that the folder structure within a zipped EPUB may vary from one file to another. 
The file paths could differ when dealing with different EPUB files.

### DIFF
--- a/src/pages/reader/epub.vue
+++ b/src/pages/reader/epub.vue
@@ -96,6 +96,7 @@ export default {
   data() {
     return {
       toc: [],
+      dirPath: '',
       selectedChapter: '',
       book: null,
       rendition: null,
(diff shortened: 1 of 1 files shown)


src/pages/reader/epub.vue
```diff
@@ -129,7 +129,10 @@ export default {
       });
       this.rendition.display();
       this.rendition.on('rendered', (cfiRange, contents) => {
-        this.selectedChapter = this.rendition.currentLocation().start.href;
+        const path = this.rendition.currentLocation().start.href;
+        const pathArr = path.split('/');
+        this.selectedChapter = pathArr.pop();
+        this.dirPath = pathArr.join('/');
```
williamchong, line 135: `dirPath` should be defined in data, and maybe cleared on reload or unmount

WeiJun0827: missed one line when committing...
`dirPath` and `selectedChapter` will be updated when rendering

williamchong (approved):

nwingt (approved):

PR #1494, opened 2024-01-11
🐛 Fix class id list was directly used in prefetch header and json ld

williamchong:
`classId` happens to be `string|string[]`

### DIFF
--- a/src/pages/store/index.vue
+++ b/src/pages/store/index.vue
@@ -91,25 +91,28 @@ export default {
     const title = this.$t('store_index_page_title');
     const description = this.$t('store_books_page_description');
     const link = [{ rel: 'canonical', href: `${this.$route.path}` }];
-    this.nftBooks.forEach(nft => {
+    const classIds = Array.from(
+      new Set(this.nftBooks.map(b => b.classId).flat())
+    );
+    classIds.forEach(classId =>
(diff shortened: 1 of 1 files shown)


nwingt (approved):

src/pages/store/index.vue
```diff
@@ -91,25 +91,35 @@ export default {
     const title = this.$t('store_index_page_title');
     const description = this.$t('store_books_page_description');
     const link = [{ rel: 'canonical', href: `${this.$route.path}` }];
-    this.nftBooks.forEach(nft => {
+    const classIds = this.nftBooks.reduce((acc, cur) => {
+      if (Array.isArray(cur.classId)) {
+        return [...acc, cur.classId];
+      }
+      if (cur.classId.includes(',')) {
+        const ids = cur.classId.split(',').map(id => id.trim());
```
nwingt, line 99: I can confirm that the class ID is either a string that doesn't need to be split by `','`, or an array. However, this is also acceptable.

nwingt (approved):

src/pages/store/index.vue
williamchong: maybe i should just use flat

nwingt (approved):

PR #1500, opened 2024-01-12
🚚 Move /writing-nft/about to /about/writing-nft

nwingt:


src/pages/writing-nft/about.vue
```diff
@@ -1,75 +1,7 @@
...
-              })
-            ),
-          }),
-          type: 'application/ld+json',
-          body: true,
-        },
-      ],
-    };
+  fetch({ redirect, localeLocation, query }) {
+    redirect(302, localeLocation({ name: 'about-writing-nft', query }));
```
williamchong, line 4: Please use 301 in this case

nwingt: Updated

PR #1506, opened 2024-01-16
[TAS-1064] 🥅 Guard invalid claim wallet address

AuroraHuang22:
截圖 2024-01-16 下午5 44 05


nwingt (changes requested):
Can we also block the Next button when there is an error?

williamchong (changes requested):
Should also block next button in onEnterClaimingAddress or with disable

williamchong (approved):

nwingt (approved):

PR #1508, opened 2024-01-16
✨ Support collection in gift page

williamchong:


src/pages/nft/gift/index.vue
```diff
@@ -4,7 +4,15 @@
...
-            localeLocation({ name: 'nft-class-classId', params: { classId } })
+            collectionId
+              ? localeLocation({
+                  name: 'nft-collection-collectionId',
+                  params: { collectionId },
+                })
+              : localeLocation({
+                  name: 'nft-class-classId',
+                  params: { classId },
+                })
```
nwingt, line 15: Can be put this in `nftOrCollectionMixin` and reuse it in `handleClickViewClass`?

src/pages/nft/gift/index.vue
```diff
@@ -91,6 +104,12 @@ export default {
     classId() {
       return this.$route.query.class_id;
     },
+    collectionId() {
+      return this.$route.query.collection_id;
+    },
+    primaryKey() {
+      return this.collectionId || this.classId;
+    },
```
nwingt, line 112: This should put in `nftOrCollectionMixin`?

src/pages/nft/gift/index.vue
williamchong: added but not sure how to name this

src/pages/nft/gift/index.vue
williamchong: yes in fact it already exists in there...

src/pages/nft/gift/index.vue
nwingt: I think is fine for now

nwingt (approved):

PR #1531, opened 2024-01-29
[TAS-1042] 🚸 Save epub reader current page to localstorage

williamchong:


WeiJun0827 (approved):

src/pages/reader/epub.vue
```diff
@@ -116,10 +123,16 @@ export default {
...
   methods: {
-    initRendition() {
+    async initRendition() {
+      this.isLoading = true;
       const encodedUrl = encodeURIComponent(this.fileSrc);
       const corsUrl = `https://pdf-cors-ufdrogmd2q-uw.a.run.app/pdf-cors?url=${encodedUrl}`;
-      this.book = Epub(corsUrl);
+      const buffer = await this.$axios.$get(corsUrl, {
+        responseType: 'arraybuffer',
+      });
```
nwingt, line 132: Should we add try-catch here?

williamchong: If any error is thrown here, the whole reader fail to function
Maybe we should show some error message

nwingt: It is acceptable if it shows an error page with the raw message

nwingt (approved):

src/pages/reader/epub.vue
williamchong: added `this.$nuxt.error`

PR #1541, opened 2024-02-20
🐛 Fix broken default nft and collection image

williamchong:
turns out webpack cannot handle ternary operation in this case


nwingt (approved):

AuroraHuang22 (approved):

williamchong:
> But local dev works?

Doesn't work, but not obvious when it's empty portfolio

PR #1542, opened 2024-02-20
🐛 Do not use nft mixin in non nft class component

williamchong:
- `NFTPortfolio/MainView.vue` contains many nft class instead of 1 nft class, but nft mixin expect an nft class id as it's root of data source
- We can't call `fetchRecommendInfo` from here, `iscnOwner` is undefined
- Somehow `portfolioItemsTrimmed` is modified if we call `fetchRecommendInfo` under these condition causing `fetchRecommendInfo` goes into an infinite loop

For a very quick fix, we should just not call `fetchRecommendInfo` from here, and maybe figure out why `portfolioItemsTrimmed` is modified and call `this.fetchCollectedNFTClassesByAddress(this.getAddress)` in a later PR


williamchong:
> So recommended info in a blank portfolio will be empty now?

No, default feature book should be used if no info is fetched

AuroraHuang22 (approved):

PR #1551, opened 2024-02-29
🚸 Fix paid & free page initial scroll position

nwingt:


src/pages/store.vue
```diff
@@ -229,10 +229,14 @@ export default {
...
-    // since it will trigger when navigating to different child components
-    // Ref: https://v2.nuxt.com/docs/components-glossary/scrolltotop#the-scrolltotop-property
-    window.scrollTo(0, 0);
+    if (this.currentTab === 'featured') {
+      // scrollToTop: true in Nuxt component is buggy in this case
+      // since it will trigger when navigating to different child components
+      // Ref: https://v2.nuxt.com/docs/components-glossary/scrolltotop#the-scrolltotop-property
+      window.scrollTo(0, 0);
+    } else {
+      this.$refs.tabsSection.scrollIntoView();
```
williamchong, line 238: Do we need to guard if this.$refs.tabsSection exists?

nwingt: we can guard it but I don't think that is necessary

williamchong (approved):

PR #1552, opened 2024-02-29
🐛 Fix mixin conflict

williamchong:
`nft` mixin shadowed `nft-or-collection` mixin, should change order

### DIFF
--- a/src/mixins/nft.js
+++ b/src/mixins/nft.js
@@ -1427,8 +1427,8 @@ export default {
     async fetchRecommendInfo() {
       if (this.isRecommendationLoading) return;
       this.isRecommendationLoading = true;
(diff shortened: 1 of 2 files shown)


src/mixins/nft.js
```diff
@@ -1427,11 +1427,12 @@ export default {
     async fetchRecommendInfo() {
       if (this.isRecommendationLoading) return;
       this.isRecommendationLoading = true;
+      const promises = [];
```
nwingt, line 1430: Why move this out from try-catch block?

williamchong: agree that inside seems better

nwingt (approved):

PR #1562, opened 2024-03-11
🐛 Fix route to undefined path when beginning checkout

williamchong:
Actually I have no idea when `handleGotoCollectFromControlBar` exists after all the checks for stock without checking nftId exists
please check if my fix makes sense


nwingt (approved):

nwingt:
@AuroraHuang22 Please confirm
image

PR #1564, opened 2024-03-13
[TAS-1435] ⚡️ Lazy load book store info

williamchong:


src/pages/nft/class/_classId/index.vue
```diff
@@ -459,16 +459,14 @@ export default {
...
             }
           }),
-      ]);
-      if (metadata?.nft_meta_collection_id?.includes('nft_book')) {
-        await store
-          .dispatch('fetchNFTBookInfoByClassId', classId)
+        store
+          .dispatch('lazyFetchNFTBookInfoByClassId', classId)
           .catch(error => {
             if (error.response?.status !== 400) {
```
nwingt, line 465: Is there a case that returns 404?

williamchong: The API now returns 400 for cases that it should return 404, should make another PR pair to "fix" it

nwingt: Should we include 404 here, so this will not break when the API is fixed?

nwingt (approved):

PR #1569, opened 2024-03-14
⚡️ Remove service worker and cache

williamchong:


nwingt (approved):

williamchong:
@nwingt Rewrote using try catch, please review again thanks

nwingt (approved):

PR #1572, opened 2024-03-14
💄 Fix missing viewport meta

williamchong:


nwingt (approved):
I thought Nuxt included this by default, what was broken?

williamchong:
Probably related to pwa module being removed

PR #1594, opened 2024-04-17
[TAS-1565] 👽️ Swap pdf proxy function to static.like.co

williamchong:


nwingt (approved):

williamchong:
@nwingt Seems firebase hosting has issue for large file and the file would be corrupted, sad

PR #1597, opened 2024-04-19
[TAS-1148] ⚡️ Use cache API to cache epub file

williamchong:
https://developer.mozilla.org/en-US/docs/Web/API/Cache
Experimental, but this API is widely supported and has large size quota compared to localStorage

TODO: purge unused key according to certain logic on mount

### DIFF
--- a/src/pages/reader/epub.vue
+++ b/src/pages/reader/epub.vue
@@ -129,9 +129,24 @@ export default {
         this.isLoading = true;
         const encodedUrl = encodeURIComponent(this.fileSrc);
         const corsUrl = `https://static2.like.co/pdf-cors/?url=${encodedUrl}`;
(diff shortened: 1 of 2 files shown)


src/pages/reader/epub.vue
```diff
@@ -129,9 +129,24 @@ export default {
...
-        const buffer = await this.$axios.$get(corsUrl, {
-          responseType: 'arraybuffer',
-        });
+        let buffer;
+        if (window.caches) {
+          try {
+            const cache = await caches.open('reader-epub');
+            let response = await cache.match(corsUrl);
+            if (!response) response = await cache.add(corsUrl);
+            console.log(response);
```
nwingt, line 138:
```suggestion
```

nwingt (changes requested):

PR #1600, opened 2024-04-22
[TAS-1568] ⚡️ Use cache API and postMessage for pdf books

williamchong:
Combines with:
https://github.com/likecoin/pdf.js/pull/4
https://github.com/likecoin/pdf.js/pull/5

### DIFF
--- a/src/pages/reader/pdf.vue
+++ b/src/pages/reader/pdf.vue
@@ -1,9 +1,24 @@
 <template>
-  <iframe :src="pdfIframeSrc" width="100%" height="100%" style="border:none" />
+  <div class="w-full h-full">
+    <div
+      v-if="isLoading"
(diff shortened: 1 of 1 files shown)


src/pages/reader/pdf.vue
```diff
@@ -14,18 +29,103 @@ export default {
...
+        if (this.isIframeReady) {
+          this.$refs.pdfIframe.contentWindow.postMessage(
+            JSON.stringify({
+              action: 'openBase64File',
+              data: {
+                data: this.base64FileData,
+                name: this.nftName,
+              },
+            }),
+            this.iframeOrigin
```
nwingt, line 115: Can be refactored

nwingt (approved):

PR #1607, opened 2024-05-07
[TAS-1627] 🐛 Fix missing bookstore info when claiming free books

AuroraHuang22:


src/pages/nft/claim/index.vue
```diff
@@ -739,15 +742,38 @@ export default {
...
+            classId: this.classId,
+            collectionId: this.collectionId,
+            paymentId: this.paymentId,
+          })
+        );
+        ({ price } = data);
+        const {
+          giftInfo,
+          isPhysicalOnly,
+          status,
```
williamchong, line 758: Why there is `giftInfo` and `status` in a class ID price data? What are their expected value?

williamchong (changes requested):

src/pages/nft/claim/index.vue
```diff
@@ -739,15 +742,38 @@ export default {
...
+        }
+      } catch (err) {
+        // eslint-disable-next-line no-console
+        console.error(err);
+      }
+    } else {
+      const { prices } = this.getNFTBookStoreInfoByClassId(this.classId);
+      const data = prices[this.priceIndex];
       const {
         giftInfo,
```
williamchong, line 778: Why we expect `giftInfo` and `status` exists in `NFTBookStoreInfoByClassId`? What are their expected values?

williamchong (changes requested):

src/pages/nft/claim/index.vue
AuroraHuang22: I forgot that free books do not have the gifting feature, thanks 😝

williamchong (approved):

PR #1614, opened 2024-05-10
[TAS-1626] 👽️ Use new ebook-cors API for reader

williamchong:


src/pages/reader/pdf.vue
```diff
@@ -75,28 +87,30 @@ export default {
         }
       }
     },
+    async getFileBuffer() {
+      let buffer;
+      if (window.caches) {
+        try {
+          const cache = await caches.open('reader-epub');
```
nwingt, line 94: PDF uses `reader-epub` too, how about a generic name for the cache? e.g. `reader-ebook`?

src/pages/reader/pdf.vue
```diff
@@ -75,28 +87,30 @@ export default {
...
+          console.error(error);
+        }
+      }
+      if (!buffer) {
+        buffer = await this.$axios.$get(this.corsUrl, {
+          responseType: 'arraybuffer',
+        });
+      }
+      return buffer;
+    },
```
nwingt, line 109: Refactor with epub?

src/pages/reader/pdf.vue
williamchong: actually it is suppose to be `reader-pdf`...

src/pages/reader/pdf.vue
```diff
@@ -18,16 +18,29 @@
...
     },
+    corsUrl: {
+      type: String,
+      default: '',
+    },
+    cacheKey: {
+      type: String,
+      default: '',
+    },
   },
```
nwingt, line 44: Props can be moved to reader mixin too

nwingt (approved):

PR #1616, opened 2024-05-14
🐛 Fix automatic redirection after login

AuroraHuang22:
Avoid the situation where you are already logged in but not redirected.
截圖 2024-05-14 上午10 27 10

### DIFF
--- a/src/pages/nft/claim/index.vue
+++ b/src/pages/nft/claim/index.vue
@@ -737,11 +737,11 @@ export default {
       }
     },
     state(newValue) {
-      if (newValue) {
-        this.$router.push({ query: { ...this.$route.query, state: newValue } });
-      }
       if (newValue === NFT_CLAIM_STATE.LOGIN && this.claimingAddress) {
         this.state = NFT_CLAIM_STATE.ID_CONFIRMATION;
+        this.navigateToState(NFT_CLAIM_STATE.ID_CONFIRMATION);
+      } else if (newValue) {
+        this.navigateToState(newValue);
       }
     },
   },
@@ -1035,12 +1035,6 @@ export default {
         this.primaryKey,
         1
       );
-      console.log(
-        'this.state',
-        this.state,
-        'this.claimingAddress',
-        this.claimingAddress
-      );
       switch (this.state) {
         case NFT_CLAIM_STATE.WELCOME:
           if (this.claimingAddress) {
@@ -1169,6 +1163,9 @@ export default {
         1
       );
       this.$router.push(this.viewInfoLocation);
+    },
+    navigateToState(state) {
+      this.$router.push({ query: { ...this.$route.query, state } });
     },
   },
 };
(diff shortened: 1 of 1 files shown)


src/pages/nft/claim/index.vue
```diff
@@ -1029,6 +1035,12 @@ export default {
         this.primaryKey,
         1
       );
+      console.log(
```
nwingt, line 1038: Please remove this log

src/pages/nft/claim/index.vue
```diff
@@ -737,6 +740,9 @@ export default {
       if (newValue) {
         this.$router.push({ query: { ...this.$route.query, state: newValue } });
       }
+      if (newValue === NFT_CLAIM_STATE.LOGIN && this.claimingAddress) {
+        this.state = NFT_CLAIM_STATE.ID_CONFIRMATION;
+      }
```
nwingt, line 745: This should be put before `$router.push()` if check?

AuroraHuang22: [🎨 Set state before pushing query](https://github.com/likecoin/liker-land/pull/1616/commits/98883bf5e71c03370eb450499af192294ebee1fd)

src/pages/nft/claim/index.vue
```diff
@@ -729,13 +729,19 @@ export default {
     },
     loginAddress(newValue) {
       this.claimingAddress = newValue;
+      if (this.claimingAddress && this.state === NFT_CLAIM_STATE.LOGIN) {
+        this.state = NFT_CLAIM_STATE.ID_CONFIRMATION;
+      }
       if (!this.getAddress && !(this.status === 'completed')) {
         this.claimingAddress = '';
       }
```
williamchong, line 737: Would the `this.getAddress` logic below affect how we judge if `this.claimingAddress` should trigger state change?

src/pages/nft/claim/index.vue
```diff
@@ -729,13 +729,19 @@ export default {
...
     },
     state(newValue) {
-      if (newValue) {
-        this.$router.push({ query: { ...this.$route.query, state: newValue } });
+      if (newValue === NFT_CLAIM_STATE.LOGIN && this.claimingAddress) {
+        this.state = NFT_CLAIM_STATE.ID_CONFIRMATION;
+        this.navigateToState(NFT_CLAIM_STATE.ID_CONFIRMATION);
+      } else if (newValue) {
```
williamchong, line 743: It seems that this logic could be done in `navigateToState` instead of a watcher, if it's generic to any state change

src/pages/nft/claim/index.vue
AuroraHuang22: Removed, since the loginAddress is stored in session storage, we can just determine if a user is logged out by checking getAddress.

src/pages/nft/claim/index.vue
AuroraHuang22: [🎨 Utilize navigateToState for all state management](https://github.com/likecoin/liker-land/pull/1616/commits/0e26cf2d3f56535f180a83b571367a852b85e796)

williamchong (approved):
It seems that we should stick with using just one of `getAddress` or `loginAddress` in the future, and dont use different logic for two

PR #1621, opened 2024-05-23
[TAS-1633] 💄 Update tipping select UI

AuroraHuang22:
截圖 2024-05-23 上午11 41 04
截圖 2024-05-23 上午11 41 08


src/components/NFTBook/TippingDialog.vue
```diff
@@ -50,16 +58,54 @@
...
+                'rounded-[50%]',
+                'bg-like-green',
+                'border-[1px]',
+                'border-white',
+                'absolute',
+                'top-[-10px]',
+                'left-[-10px]',
+              ]"
+            >
+              <IconCheckMini class="text-white" />
```
nwingt, line 78: How about making the SVG absolute and removing the parent div

src/components/NFTBook/TippingDialog.vue
```diff
@@ -50,16 +58,54 @@
...
+              'bg-like-green',
+              'border-[1px]',
+              'border-white',
+              'absolute',
+              'top-[-10px]',
+              'left-[-10px]',
+            ]"
+          >
+            <IconCheckMini class="text-white" />
+          </div>
```
nwingt, line 108: Seems can be refactorized

AuroraHuang22: [♻️ Refactor check icon](https://github.com/likecoin/liker-land/pull/1621/commits/1bf6a7014400f7febed25c82b71fa09ce31bc088)

src/components/NFTBook/TippingDialog.vue
AuroraHuang22: The parent div contains a circular background

PR #1623, opened 2024-05-27
🥅 Catch contentWindow is undefined

williamchong:
Strange error when opening pdf for first time, haven't have time to investigate yet


nwingt:
How about listen to some kind of on load event from the iframe and post message after that?

williamchong:
> How about listen to some kind of on load event from the iframe and post message after that?

I'm already doing that, but we probably still need to handle two cases, where the load event happen sooner or later then file download

PR #1632, opened 2024-06-13
💄 Add ellipsis to Identity

AuroraHuang22:
截圖 2024-06-13 下午4 14 43

### DIFF
--- a/src/components/NFTPortfolio/MainView.vue
+++ b/src/components/NFTPortfolio/MainView.vue
@@ -213,6 +213,8 @@
 </template>
 
 <script>
+import { mapGetters } from 'vuex';
+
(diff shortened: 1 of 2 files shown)


src/components/NFTPortfolio/MainView.vue
```diff
@@ -307,6 +309,7 @@ export default {
     };
   },
   computed: {
+    ...mapGetters(['getAddress']),
```
nwingt, line 312: What is this for?

AuroraHuang22: It's quick fix; this page hasn't imported the mixin, resulting in 'getAddress' being undefined

nwingt (approved):

PR #1638, opened 2024-06-17
💄 Update book cover height for /claim

AuroraHuang22:
IMG_3865

截圖 2024-06-17 下午2 07 22

### DIFF
--- a/src/components/NFTClaim/MainSection.vue
+++ b/src/components/NFTClaim/MainSection.vue
@@ -6,7 +6,7 @@
 
       'px-[24px]',
       'laptop:px-[12px]',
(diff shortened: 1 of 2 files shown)


src/components/NFTClaim/MainSection.vue
```diff
@@ -6,7 +6,7 @@
 
       'px-[24px]',
       'laptop:px-[12px]',
-      'mt-[72px]',
+      'mt-[78px]',
```
nwingt, line 9: Why adding 6px? 😂

nwingt (approved):

PR #1639, opened 2024-06-17
[TAS-1770] 🔍 Display all result if no searchKeyword

AuroraHuang22:
There is a delay in the page when the last keyword is deleted, so I use debounced here.

[https://rinkeby.liker.land/zh-Hant/gutenberg/free-audio-books](https://rinkeby.liker.land/zh-Hant/gutenberg/free-audio-books)

### DIFF
--- a/src/pages/gutenberg/free-audio-books.vue
+++ b/src/pages/gutenberg/free-audio-books.vue
@@ -68,10 +68,11 @@
       >
         <IconSearch />
         <input
-          v-model="searchKeyword"
+          :value="searchKeyword"
           class="w-full bg-transparent border-0 focus-visible:outline-none"
           type="text"
           :placeholder="$t('gutenberg_search_placeholder')"
+          @input="debouncedUpdateSearchKeyword"
           @change="handleInputChange"
         />
       </div>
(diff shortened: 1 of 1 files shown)


src/pages/gutenberg/free-audio-books.vue
```diff
@@ -364,6 +350,12 @@ export default {
     handleInputChange(value) {
       logTrackerEvent(this, 'Gutenberg', 'inputChange', value.target.value, 1);
     },
+    debouncedUpdateDisplayNumber: debounce(
```
williamchong, line 353: Is it better to debounce "setting input to `searchKeyword`" instead of this number? We dont need per character update on the result

williamchong (changes requested):

src/pages/gutenberg/free-audio-books.vue
AuroraHuang22: [🎨 Use debounce for setting searchKeyword](https://github.com/likecoin/liker-land/pull/1639/commits/5569cf0c6adc94c6eba29236e53631073b015fd0)

nwingt (approved):

src/pages/gutenberg/free-audio-books.vue
```diff
@@ -314,25 +309,17 @@ export default {
...
   },
   watch: {
-    searchKeyword() {
-      this.currentDisplayNumber = DISPLAY_NUMBER;
+    searchKeyword: {
+      handler(value) {
+        if (!value) {
+          this.currentDisplayNumber = this.parsedData.length;
+        } else {
+          this.currentDisplayNumber = DISPLAY_NUMBER;
```
williamchong, line 319: Should use computed?

williamchong (approved):

PR #1642, opened 2024-06-19
[TAS-1784] ♻️ Rewrite reader fetch to avoid duplicate on cache err

williamchong:
To always use `fetch` to avoid `cache.add` error causing another xhr from axios


nwingt (approved):

williamchong:
changed to use req obj for `.match`, please help review again @nwingt

nwingt (approved):

PR #1654, opened 2024-06-21
[TAS-1790] 🐛 Fix hydration mismatch caused by html comment

williamchong:
A layout change from a hydration mismatch page would cause double mount event

### DIFF
--- a/src/pages/index.vue
+++ b/src/pages/index.vue
@@ -384,7 +384,7 @@
         <div
           class="desktop:grid grid-cols-5 items-stretch gap-[20px] flex-grow"
         >
-          <div class="col-span-2">
+          <div v-if="stickyBookstoreItem" class="col-span-2">
             <NFTBookItemCardV2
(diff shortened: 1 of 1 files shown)


nwingt:
Do you mean the mount events from `reader` and `reader/pdf`/`reader/epub` pages?

williamchong:
> Do you mean the mount events from `reader` and `reader/pdf`/`reader/epub` pages?

Epub page is mounted twice

williamchong:
Very similar to https://github.com/nuxt/nuxt/issues/5703

nwingt (approved):

williamchong:
@nwingt helmet: i haven't got anything working yet

nwingt:
Is this referencing the wrong task?

williamchong:
> Is this referencing the wrong task?

this should be the cause for both TAS-1711  and TAS-1790

nwingt (approved):
Hmmm..I left comments in other templates as well. Why is this causing an error?

PR #1682, opened 2024-07-19
[TAS-1909] 💩 Do not send likerland verify email if not verified in authcore

williamchong:


src/server/api/routes/users/v2/email.js
```diff
@@ -73,20 +74,22 @@ router.post('/email', authenticateV2Login, async (req, res, next) => {
...
-      subject: title,
-      title,
-      content: `<p>Please click the link to verify your email:</p><p>${verificationURL}</p>`,
-    });
-    await sendEmail({
-      email,
-      subject,
-      html: body,
-    });
+    if (verify !== '0') {
```
nwingt, line 77: Maybe add a comment here?

nwingt (approved):

src/server/api/routes/users/v2/email.js
williamchong: about the meaning of `verify`, or when do we set `verify === '0'` ?
Maybe we should rename `verify` if its meaning is unclear

nwingt: How about this
```suggestion
    const hasVerifiedInAuthcore = verify !== '0';
    if (hasVerifiedInAuthcore) {
```

williamchong: Renamed var and added authcore description

PR #1684, opened 2024-07-19
[TAS-1917] 💄 Revamp Epub reader UI

nwingt:
Liker Land · 02 51 · 07-20
Liker Land

### DIFF
--- a/src/pages/reader/epub.vue
+++ b/src/pages/reader/epub.vue
@@ -1,5 +1,191 @@
 <template>
-  <div>
+  <div class="fixed inset-0 flex flex-col">
+    <header class="border-b border-[#ccc]">
+      <div class="flex items-center justify-between px-[16px] py-[8px]">
+        <div class="flex items-center gap-4">
+          <NuxtLink
+            :to="
+              localeLocation({
+                name: 'dashboard',
+                query: {
+                  view: 'collectibles',
+                  tab: 'collected',
+                  type: 'nft_book',
+                },
+              })
+            "
+            :alt="$t('main_menu_my_portfolio')"
+          >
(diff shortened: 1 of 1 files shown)


src/pages/reader/epub.vue
```diff
@@ -248,7 +359,8 @@ export default {
       }
     },
     async searchEntireBook() {
-      if (this.searchText.length < 3) return [];
+      if (!this.searchText.length) return [];
```
williamchong, line 362: I think this magic number here is to avoid single character search hanging the whole page, maybe we need debounce and limit search result if we remove it

williamchong: suggest to keep the  `< 3` condition if we don't want to handle this for now
ideally we check `< 3` for alphanumeric char and skip the check for non ascii char

williamchong (approved):

src/pages/reader/epub.vue
```diff
@@ -233,155 +406,183 @@ export default {
...
       const selection = this.contents.window.getSelection();
-      // TODO: selection is null when the page is not rendered (e.g. change chapter)
+      // Selection becomes null when changing chapter
       if (selection) {
         selection.removeAllRanges();
         selection.addRange(range);
+      } else {
+        // Focus again after changing chapter
+        this.$nextTick(() => {
+          this.focusSelectedSearchResult();
```
williamchong, line 501: Would this cause infinite loop somehow though

nwingt: Use highlight instead selection for selected search result

AuroraHuang22 (approved):

PR #1685, opened 2024-07-19
👔 Preserve from channel when changing route

nwingt:


src/middleware/from-channel.js
```diff
@@ -0,0 +1,23 @@
...
+    redirect,
+    getRouteBaseName,
+  } = context;
+
+  // If the from query is present in the current route but not in the next route, add it to the next route
+  if (
+    fromRoute.query.from &&
+    !toRoute.query.from &&
+    // Exclude navigating from /nft/* routes
+    !getRouteBaseName(fromRoute).startsWith('nft')
```
williamchong, line 16: If we set applicable condition here, I think this middleware can continue to be global

williamchong (approved):

PR #1692, opened 2024-07-29
👔 Do not expose URL for downloadable book

nwingt:


williamchong (approved):
Should review where contentUrl is being used now, seems it doesn't really have use case for book

PR #1694, opened 2024-08-01
💄 Remove epub reader width limit

nwingt:
Seems viewer's width affects images' height, removing width limit to viewer can mitigate the spacing issue

Before
image

After
image

### DIFF
--- a/src/pages/reader/epub.vue
+++ b/src/pages/reader/epub.vue
@@ -142,25 +142,13 @@
       </div>
     </header>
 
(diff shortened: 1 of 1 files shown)


AuroraHuang22 (approved):

williamchong (approved):
Troublesome that the epub css interacts with our site css, not sure if we should stuff it in iframe or some container

nwingt:
I think it is related to epubjs, our CSS does not affected the viewer since it is already inside an iframe
image

PR #1697, opened 2024-08-02
[TAS-1947] 📈 Fix ecommerce event in carts and support quantity

williamchong:


src/util/EventLogger.js
```diff
@@ -94,14 +95,17 @@ export function logPurchaseFlowEvent(
...
-        items: items.map(i => ({
-          item_id: i.classId,
-          item_name: i.name?.substring(0, 100),
-          item_brand: isNFTBook ? 'Book NFT' : 'Writing NFT',
-          currency,
-          price: i.price,
-          quantity: 1,
-        })),
+        items: items.map(i => {
+          const itemId = i.productId || i.collectionId || i.collectionId;
```
nwingt, line 99:
```suggestion
          const itemId = i.productId || i.collectionId;
```

src/pages/shopping-cart/wnft.vue
```diff
@@ -207,6 +207,7 @@ export default {
   mounted() {
     this.updateLIKEPrice();
     logPurchaseFlowEvent(this, 'add_to_cart', this.purchaseEventParams);
```
nwingt, line 209: Should we keep triggering `add_to_cart` when entering this page?

src/pages/nft/collection/_collectionId/index.vue
```diff
@@ -287,20 +298,29 @@ export default {
...
-                price: this.collectionPrice,
+                price: customPriceInDecimal || this.collectionPrice,
                 collection: this.collectionId,
               },
             ],
-            price: this.collectionPrice,
+            price: totalPrice,
             currency: 'USD',
             isNFTBook: true,
           };
```
nwingt, line 317: Is this replaced by `this.purchaseEventParams`?

src/pages/shopping-cart/wnft.vue
williamchong: In wnft I think the user would always go into cart page when clicking `purchase all`?
But wnft flow is not very important now i guess

src/pages/nft/collection/_collectionId/index.vue
williamchong: Yes

nwingt (approved):

PR #1701, opened 2024-08-04
🐛 Fix index field is not used in book price

williamchong:
Related to https://github.com/likecoin/nft-book-press/issues/156#issuecomment-2266338821

### DIFF
--- a/src/pages/nft/class/_classId/index.vue
+++ b/src/pages/nft/class/_classId/index.vue
@@ -1028,10 +1028,16 @@ export default {
       );
       return this.handleCollect();
     },
+    getEdition(selectedValue) {
+      const editions = this.getNFTBookStorePricesByClassId(this.classId) || {};
+      const index = selectedValue ?? this.selectedValue;
+      const edition =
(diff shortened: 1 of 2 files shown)


src/pages/nft/class/_classId/index.vue
```diff
@@ -1084,7 +1088,9 @@ export default {
     },
     async handleCollectFromEdition(selectedValue, giftInfo = undefined) {
       const editions = this.getNFTBookStorePricesByClassId(this.classId) || {};
-      const edition = editions[selectedValue ?? this.selectedValue];
+      const index = selectedValue ?? this.selectedValue;
+      const edition =
+        editions.find(e => e.index === Number(index)) || editions[index];
```
nwingt, line 1093: Can be refactored?

williamchong: Actually I'm not sure if selectedValue must be number, if yes this can be simpler

nwingt (approved):

PR #1722, opened 2024-08-15
[TAS-2016] 🔧 Allow scripted content for ePub reader

nwingt:


williamchong (approved):
Will use iframe + book3app in the future

williamchong (changes requested):
@nwingt on second thought, we actually hard code a list of class id that can run script instead of enabling for all

nwingt:
How about owner wallet?

williamchong:
> How about owner wallet?

Would also work but unless aome wallet always has interactive book, class id is safer

williamchong (approved):

PR #1728, opened 2024-08-21
[TAS-2045] 📈 Add tipping related events

williamchong:


src/components/NFTBook/TippingDialog.vue
```diff
@@ -171,6 +171,7 @@ export default {
     },
   },
   mounted() {
+    logTrackerEvent(this, 'NFT', 'NFTTippingOpen', this.currentId, 1);
```
nwingt, line 174: is this a new event? probably use snake case for new event

williamchong: To be consistent with existing event in thos component, otherwise it is hard to filter

nwingt (approved):

PR #1741, opened 2024-08-26
[TAS-1896] 🚸 Add nocache query to fetchCollectedNFTClassesByAddress

AuroraHuang22:


src/util/api/index.js
```diff
@@ -177,6 +178,7 @@ export const getNFTClassesPartial = ({
   if (reverse) qsPayload['pagination.reverse'] = reverse;
   if (limit) qsPayload['pagination.limit'] = limit;
   if (key) qsPayload['pagination.key'] = key;
+  if (nocache) qsPayload.nocache = `${new Date().getTime()}`;
```
williamchong, line 181:
```suggestion
  if (nocache) qsPayload.ts = `${Math.round(new Date().getTime() / 1000)}`;
```

just call the ts ts seems simpler, also maybe round to second?

src/util/nft.js
```diff
@@ -298,13 +298,20 @@ export function formatNFTClassInfo(classData) {
...
   axios,
-  { iscnOwner, nftOwner }
+  { iscnOwner, nftOwner, nocache = false }
 ) => {
+  const params = {
+    iscnOwner,
+    nftOwner,
+  };
+  if (nocache) {
+    params.nocache = nocache;
```
williamchong, line 308: seems the behaviour would be the same if we just pass whole params in with `nocache` no matter true or false?

nwingt (approved):

williamchong (approved):

PR #1745, opened 2024-08-26
🚸 Skip login if login address exist

nwingt:


src/pages/nft/claim/index.vue
```diff
@@ -548,7 +548,7 @@ export default {
   data() {
     return {
       nftId: '',
-      state: NFT_CLAIM_STATE.WELCOME,
+      state: this.$route.query.state || NFT_CLAIM_STATE.WELCOME,
```
williamchong, line 551: Not sure if we need to check if state qs is valid

src/pages/nft/claim/index.vue
```diff
@@ -890,6 +890,8 @@ export default {
       this.navigateToState(NFT_CLAIM_STATE.ERROR);
     } else if (state === NFT_CLAIM_STATE.LOGIN && this.claimingAddress) {
       this.navigateToState(NFT_CLAIM_STATE.ID_CONFIRMATION);
+    } else if (this.loginAddress) {
```
williamchong, line 893: Can we merge these 2 condition `state === NFT_CLAIM_STATE.LOGIN && this.claimingAddress` and  `this.loginAddress` ?

williamchong (approved):

PR #1756, opened 2024-08-30
[TAS-2134] 🐛 Fix email is empty in free purchase

nwingt:
In `data()`, `walletEmail` is `undefined`
https://github.com/likecoin/liker-land/blob/52a1737950ede7fd19de0ffe139e35f2002eadd4/src/pages/nft/claim/index.vue#L548-L559

Screenshot 2024-08-30 at 15 55 25

### DIFF
--- a/src/pages/nft/claim/index.vue
+++ b/src/pages/nft/claim/index.vue
@@ -924,6 +924,7 @@ export default {
     async startFreePurchase() {
       try {
         this.isClaimLoading = true;
+        this.claimingFreeEmail = this.walletEmail;
         this.navigateToState(NFT_CLAIM_STATE.CLAIMING);
         if (!this.claimingFreeEmail && !this.claimingAddress) {
           this.alertPromptError(


src/pages/nft/claim/index.vue
```diff
@@ -709,6 +709,8 @@ export default {
     },
   },
   async mounted() {
+    this.claimingFreeEmail = this.walletEmail;
```
williamchong, line 712: set only if `free` is set? Not sure if we have such if case below for use

nwingt: Moved to `startFreePurchase()`

williamchong (approved):

AuroraHuang22 (approved):

src/pages/nft/claim/index.vue
williamchong: Why did the watcher on `walletEmail` didn't work though

nwingt: my case is already logged in class page

PR #1762, opened 2024-09-02
[TAS-2145] 📈 Set and reset crisp session on login

williamchong:


src/util/EventLogger.js
```diff
@@ -136,10 +166,32 @@ export function logPurchaseFlowEvent(
...
+              {
+                price,
+                currency,
+                items: items.map(
+                  i => i.productId || i.collectionId || i.classId
+                ),
+              },
+            ],
+          ],
+        ],
```
nwingt, line 192: 3 nested arrays..., is it intentional?

src/util/EventLogger.js
```diff
@@ -51,16 +66,24 @@ export async function setLoggerUser(
   }
 }
 
-export function updateSentryUser(vue, { user, displayName }) {
-  if (user) {
+export function updateUserInfo(vue, { wallet, displayName, email }) {
```
nwingt, line 69: Suggest to name this to `updateLoggerUserInfo()`

AuroraHuang22 (approved):

src/util/EventLogger.js
williamchong: Yes... https://docs.crisp.chat/guides/chatbox-sdks/web-sdk/dollar-crisp/#push-session-events

PR #1764, opened 2024-09-03
💬 Update translation

AuroraHuang22:


nwingt:
Can we remove `：` too

PR #1766, opened 2024-09-03
✨ Support cross-selling for free product

nwingt:


src/pages/nft/claim/index.vue
```diff
@@ -707,6 +722,9 @@ export default {
         this.navigateToState(NFT_CLAIM_STATE.ID_CONFIRMATION);
       }
     },
+    $route() {
+      this.showCrossSellDialogIfNecessary();
```
williamchong, line 726: What are we watching for here?

PR #1774, opened 2024-09-04
📱 Add option list for mobile UI in /claim

AuroraHuang22:
https://github.com/user-attachments/assets/7d0ca19e-0c2d-4a35-b477-ad128c72635f


nwingt:
I think we can replace all view option lists for mobile, not just the one on claim page

Should make the existing one responsive instead of creating new one

nwingt (approved):

PR #1775, opened 2024-09-04
[TAS-2146] ✨ Add wallet query to wallet api

williamchong:


src/server/api/routes/users/v2/wallet.js
```diff
@@ -7,25 +7,43 @@ const router = Router();
...
+    } else {
+      res.status(400).send('MISSING_QUERY');
       return;
     }
-    const wallet = userDoc.docs[0].id;
-    res.json({ wallet });
+    const userWallet = userDoc.id;
+    const { lastLoginMethod, registerLoginMethod } = userDoc.data();
+    res.json({
+      wallet: userWallet,
```
nwingt, line 43: should we return email as well?

nwingt (approved):

PR #1777, opened 2024-09-05
[TAS-2182] 🚸 Display logged in email in claim page

nwingt:
image


williamchong (approved):
Using current color/font, maybe it would be better if we do

Name `wallet`
email

 but current version is good enough

PR #1779, opened 2024-09-05
💄 Remove STHeiti font

nwingt:
STHeiti don't have support font-weight

Before
image

After
image


williamchong (approved):
`Before` look cleaner to me though

PR #1782, opened 2024-09-09
[TAS-2157] 📈 Store utm info to session storage

williamchong:


src/mixins/utm.js
```diff
@@ -1,14 +1,26 @@
+import { setSessionStorageItem, getFromSessionStorage } from '@/util/misc';
+
+const INTERNAL_REFERRERS = [
+  'https://liker.land/',
+  'https://liker.land/en',
+  'https://liker.land/zh-Hant',
```
nwingt, line 6: Since we use `.includes`, these can be combined to `https://liker.land/`

nwingt (approved):

PR #1802, opened 2024-09-13
[TAS-2223] 🔊 Publish user api log

williamchong:


src/server/api/routes/users/v2/email.js
```diff
@@ -155,7 +166,7 @@ router.put('/email', async (req, res, next) => {
       if (!userDoc.exists) {
         throw new Error('USER_NOT_FOUND');
       }
-      const { emailUnconfirmed: email, emailVerifyToken } = userDoc.data();
+      const { emailUnconfirmed, emailVerifyToken } = userDoc.data();
```
nwingt, line 169: What about the email in the payload?

williamchong: This kind of shadowing is so hard to handle..

nwingt (approved):

PR #1820, opened 2024-09-19
[TAS-2244] 🐛Update latest ownerInfo for Item card

AuroraHuang22:
`this.lazyFetchNFTClassAggregatedData()` resets the legacy owner information, setting `ownerCount` to 0.
When `ownerCount` is 0, the `open book` button will not be displayed.

### DIFF
--- a/src/components/NFTPortfolio/Item.vue
+++ b/src/components/NFTPortfolio/Item.vue
@@ -102,6 +102,7 @@ export default {
   methods: {
     fetchInfo() {
       this.lazyFetchNFTClassAggregatedData();
+      this.updateNFTOwners();
     },
     async handleClickCollect() {
       this.$emit('collect');
--- a/src/components/NFTPortfolio/Item.vue
+++ b/src/components/NFTPortfolio/Item.vue
@@ -100,8 +100,9 @@ export default {
     },
   },
   methods: {
-    fetchInfo() {
-      this.lazyFetchNFTClassAggregatedData(['owner']);
+    async fetchInfo() {
+      await this.lazyFetchNFTClassAggregatedData(['owner']);
+      this.updateNFTOwners();
     },
     async handleClickCollect() {
       this.$emit('collect');
--- a/src/components/NFTPortfolio/Item.vue
+++ b/src/components/NFTPortfolio/Item.vue
@@ -101,8 +101,7 @@ export default {
   },
   methods: {
     fetchInfo() {
-      this.lazyFetchNFTClassAggregatedData();
-      this.updateNFTOwners();
+      this.lazyFetchNFTClassAggregatedData(['owner']);
     },
     async handleClickCollect() {
       this.$emit('collect');
--- a/src/components/NFTPortfolio/Item.vue
+++ b/src/components/NFTPortfolio/Item.vue
@@ -101,8 +101,10 @@ export default {
   },
   methods: {
     async fetchInfo() {
-      await this.lazyFetchNFTClassAggregatedData(['owner']);
-      this.updateNFTOwners();
+      await Promise.all([
+        this.lazyFetchNFTClassAggregatedData({ excludeOptions: ['owner'] }),
+        this.updateNFTOwners(),
+      ]);
     },
     async handleClickCollect() {
       this.$emit('collect');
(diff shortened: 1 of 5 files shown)


src/components/NFTPortfolio/Item.vue
```diff
@@ -102,6 +102,7 @@ export default {
   methods: {
     fetchInfo() {
       this.lazyFetchNFTClassAggregatedData();
+      this.updateNFTOwners();
```
nwingt, line 105: should we await?

williamchong:
> `this.lazyFetchNFTClassAggregatedData()` resets the legacy owner information, setting `ownerCount` to 0. When `ownerCount` is 0, the `open book` button will not be displayed.

What does `reset` and `legacy` mean here? `lazyFetchNFTClassAggregatedData` is using old cached data to overwrite `ownerCount`?

AuroraHuang22:
> > `this.lazyFetchNFTClassAggregatedData()` resets the legacy owner information, setting `ownerCount` to 0. When `ownerCount` is 0, the `open book` button will not be displayed.
> 
> What does `reset` and `legacy` mean here? `lazyFetchNFTClassAggregatedData` is using old cached data to overwrite `ownerCount`?

If we fetch the owner info right after claiming, owner info will be overwritten by the cached data when the item is mounted. Therefore, we fetch the owner info after that.

src/components/NFTPortfolio/Item.vue
AuroraHuang22: not sure if it will cause a slow loading when there are a lot of items.

williamchong: If we don't await `lazyFetchNFTClassAggregatedData` first we risk a race condition here ownerCount is overwritten again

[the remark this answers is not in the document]
williamchong: Or we should allow `lazyFetchNFTClassAggregatedData` to ignore some fields by passing parameter
- Ideally we should ignore fields by passing `data` qs to API, since the API already support this
- If we are too lazy to figure out what qs to pass to API, we can still selectively not commit data back to store according to parameters

nwingt (approved):

src/components/NFTPortfolio/Item.vue
```diff
@@ -100,8 +100,9 @@ export default {
     },
   },
   methods: {
-    fetchInfo() {
-      this.lazyFetchNFTClassAggregatedData();
+    async fetchInfo() {
+      await this.lazyFetchNFTClassAggregatedData(['owner']);
+      this.updateNFTOwners();
```
williamchong, line 105: if we exclude `owner` in `lazyFetchNFTClassAggregatedData `, then we can use `Promise.all` instead of blocking `await`

src/store/modules/nft.js
```diff
@@ -453,20 +453,25 @@ const actions = {
...
-    } = await this.$api.$get(api.getNFTClassMetadata(classId));
+    } = await this.$api.$get(api.getNFTClassMetadata(classId, excludeOptions));
     const iscnId = classData.parent.iscn_id_prefix;
     commit(TYPES.NFT_SET_NFT_CLASS_METADATA, { classId, metadata: classData });
     commit(TYPES.NFT_SET_ISCN_METADATA, { iscnId, data: iscnData });
-    commit(TYPES.NFT_SET_NFT_CLASS_OWNER_INFO, { classId, info: ownerInfo });
     commit(TYPES.NFT_SET_NFT_CLASS_LISTING_INFO, { classId, info: listings });
+    if (ownerInfo) {
+      commit(TYPES.NFT_SET_NFT_CLASS_OWNER_INFO, { classId, info: ownerInfo });
+    }
```
williamchong, line 474: should judge by `excludeOptions`, and for all fields?

src/mixins/nft.js
```diff
@@ -869,8 +874,8 @@ export default {
     lazyFetchNFTOwners() {
       return this.lazyGetNFTOwners(this.classId);
     },
-    updateNFTOwners() {
-      return this.fetchNFTOwners(this.classId);
+    updateNFTOwners({ nocache = true } = {}) {
```
williamchong, line 877: Not sure what should be the default value of `nocache`
if we don't want to change existing behaviour it should be `false`
but if we only call this in non-lazy case, `true` is more sensible

williamchong (approved):

PR #1824, opened 2024-09-20
[TAS-2267] 🚸 Allow clicking 1/3 of the epub view to navigate

nwingt:
Based on #1822


src/pages/reader/epub.vue
```diff
@@ -290,6 +292,11 @@ export default {
     this.initRendition();
     document.addEventListener('keydown', this.keyListener, false);
   },
+  beforeDestroy() {
+    if (this.cleanUpClickListener) {
+      this.cleanUpClickListener();
```
williamchong, line 297: Not sure if we need to set `this.cleanUpClickListener` to undefined here

PR #1829, opened 2024-09-23
🔒️ Set crisp token on v2 login, fix showCrisp event mess up

williamchong:
https://docs.crisp.chat/guides/chatbox-sdks/web-sdk/identity-verification/

### DIFF
--- a/src/util/EventLogger.js
+++ b/src/util/EventLogger.js
@@ -58,7 +58,7 @@ export async function setLoggerUser(
       }
     }
     if (vue.$crisp) {
-      vue.$crisp.push(['set', 'session:data', [[['wallet', wallet]]]]);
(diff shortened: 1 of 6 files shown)


src/util/EventLogger.js
```diff
@@ -77,10 +80,10 @@ export function updateLoggerUserInfo(vue, { wallet, displayName, email }) {
   }
   if (vue.$crisp) {
     if (email) {
-      vue.$crisp.push(['set', 'user:email', [email]]);
+      vue.$crisp.push(['set', 'user:email', [email, crispToken]]);
```
nwingt, line 83: When do we need to guard get user:email exist before set?

williamchong: We do not have crispToken in the other instance, so i guard to prevent overwriting verified email with unverified email

nwingt (approved):

PR #1834, opened 2024-09-24
[TAS-2258] 🐛 Fix epub reader font size inconsistent

nwingt:
image

### DIFF
--- a/src/pages/reader/epub.vue
+++ b/src/pages/reader/epub.vue
@@ -64,6 +64,32 @@
         </div>
 
         <div class="flex items-center gap-4">
+          <button @click="decreaseFontSize">
+            <svg
+              class="w-20 h-20"
+              xmlns="http://www.w3.org/2000/svg"
+              viewBox="0 0 24 24"
+            >
+              <path
+                fill="currentColor"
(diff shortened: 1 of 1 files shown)


src/pages/reader/epub.vue
```diff
@@ -347,6 +378,17 @@ export default {
         );
       }
     },
+    adjustFontSize(indexDiff) {
+      const currentIndex = FONT_SIZES.indexOf(this.fontSize);
```
williamchong, line 382: Storing fontSizeIndex as data, bound it by math.max()/min() seems more straight forward

nwingt: Updated

src/pages/reader/epub.vue
```diff
@@ -347,6 +378,20 @@ export default {
         );
       }
     },
+    adjustFontSize(indexDiff) {
+      this.fontSizeIndex = Math.max(
+        0,
+        Math.min(FONT_SIZES.length - 1, this.fontSizeIndex + indexDiff)
+      );
+      const fontSize = FONT_SIZES[this.fontSizeIndex];
+      this.rendition.themes.fontSize(`${fontSize}px`);
```
williamchong, line 387: fontSize can be computed

williamchong (approved):

PR #1851, opened 2024-09-30
📈 Set em in fbq

williamchong:


src/util/EventLogger.js
```diff
@@ -79,7 +89,17 @@ export function updateLoggerUserInfo(
     vue.$sentry.setUser(opt);
   }
   if (vue.$gtag) {
-    if (email) vue.$gtag.set('user_data', { email });
+    if (!window.doNotTrack && !navigator.doNotTrack) {
+      if (email) vue.$gtag.set('user_data', { email });
+    }
+  }
+  if (window.fbq && window.PIXEL_ID && !IS_TESTNET) {
+    if (!window.doNotTrack && !navigator.doNotTrack) {
```
nwingt, line 97: can refactor these two conditions into functions?

nwingt (approved):

PR #1852, opened 2024-09-30
🐛 Fix nftId mismatch if read owned class from non-owned nft page

williamchong:


src/mixins/nft.js
```diff
@@ -583,6 +584,13 @@ export default {
       );
       return collector?.collectedCount || 0;
     },
+    isOwningNFT() {
+      if (!this.nftId) return this.ownCount > 0;
+      const collector = this.populatedCollectors.find(
+        ({ id }) => id === this.getAddress
```
nwingt, line 590: use `loginAddress`?

nwingt (approved):

nwingt (approved):

PR #1867, opened 2024-10-07
📈 Add publish date and product group info

williamchong:
https://developers.google.com/search/docs/appearance/structured-data/product-variants

### DIFF
--- a/src/pages/nft/class/_classId/_nftId.vue
+++ b/src/pages/nft/class/_classId/_nftId.vue
@@ -502,6 +502,8 @@ export default {
           sku: this.classId,
           iscn: this.iscnId,
           isbn: this.iscnData?.contentMetadata?.isbn,
+          datePublished: this.iscnData?.recordTimestamp,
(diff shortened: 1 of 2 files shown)


nwingt (approved):

src/pages/nft/class/_classId/_nftId.vue
```diff
@@ -502,6 +502,8 @@ export default {
           sku: this.classId,
           iscn: this.iscnId,
           isbn: this.iscnData?.contentMetadata?.isbn,
+          datePublished: this.iscnData?.recordTimestamp,
+          bookFormat: this.isNftBook ? 'https://schema.org/EBook' : undefined,
```
AuroraHuang22, line 506:
```suggestion
          bookFormat: this.nftIsNFTBook ? 'https://schema.org/EBook' : undefined,
```

AuroraHuang22: Not sure if `isNftBook` should be undefined on the nftId page because it is usually passed as a prop using `nftIsNFTBook`

PR #1889, opened 2024-10-12
[TAS-2340] 🐛 Enable content viewing using ownedNFTId instead of lastOwnedNFTId

AuroraHuang22:
before
截圖 2024-10-12 下午3 13 56

after
截圖 2024-10-12 下午3 15 13

### DIFF
--- a/src/components/NFTPortfolio/Item.vue
+++ b/src/components/NFTPortfolio/Item.vue
@@ -24,7 +24,7 @@
       :display-state="nftDisplayState"
       :is-nft-book="nftIsNFTBook"
       :portfolio-tab="portfolioTab"
-      :is-content-viewable="!(nftIsNFTBook && !isOwningNFT)"
+      :is-content-viewable="isContentViewable"
       :is-content-downloadable="!nftIsDownloadHidden"
       :collect-expiry-time="collectExpiryTime"
       @collect="handleClickCollect"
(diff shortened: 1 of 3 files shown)


src/components/NFTPortfolio/Item.vue
```diff
@@ -74,8 +77,17 @@ export default {
     };
   },
   computed: {
+    ownedNftId() {
+      return this.collectorMap?.[this.getAddress]?.[0];
+    },
+    isContentViewable() {
+      return this.nftIsNFTBook && this.isBookshelf && this.ownedNftId;
```
williamchong, line 84: What if we are viewing the public view of our own portfolio?

williamchong (changes requested):

src/components/NFTPortfolio/Item.vue
AuroraHuang22: [🎨 Add current user's portfolio check](https://github.com/likecoin/liker-land/pull/1889/commits/60c5b0430007e9cf10d10d8cdde0eff58a1333f6)

src/components/NFTPortfolio/Item.vue
```diff
@@ -74,8 +77,24 @@ export default {
     };
   },
   computed: {
+    ownedNftId() {
+      return this.collectorMap?.[this.getAddress]?.[0];
+    },
```
williamchong, line 82: Seems there is a `nftIdCollectedFirstByUser` in nft mixin that we can resue

williamchong (changes requested):

src/components/NFTPortfolio/Item.vue
AuroraHuang22: [♻️ Use mixin data instead of new computed props](https://github.com/likecoin/liker-land/pull/1889/commits/f48701ae74a7ff383a0a5cb456641b7d2a5d3599)

nwingt (approved):

PR #1891, opened 2024-10-13
[TAS-2413] 📈 Set product data for 1st edition in og meta for fb

williamchong:
Since facebook cannot parse json-ld with multiple `@type` like google does, and also likely does not support product variant in a same page, we set the current edition product info in og

https://developers.facebook.com/docs/marketing-api/catalog/reference/#og-tags

### DIFF
--- a/src/pages/nft/class/_classId/index.vue
+++ b/src/pages/nft/class/_classId/index.vue
@@ -773,6 +773,11 @@ export default {
             property: 'product:condition',
             content: 'new',
           },
+          {
+            hid: 'product:custom_label_0',
+            property: 'product:custom_label_0',
+            content: this.iscnOwner,
+          },
         ].forEach(m => meta.push(m));
         meta.find(m => m.hid === 'og:url').content = `${EXTERNAL_HOST}${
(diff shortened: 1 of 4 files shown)


src/pages/nft/class/_classId/index.vue
```diff
@@ -664,34 +715,106 @@ export default {
...
+            content: this.iscnOwner,
+          },
+        ].forEach(m => meta.push(m));
+        meta.find(m => m.hid === 'og:url').content = `${EXTERNAL_HOST}${
+          this.$route.path
+        }?price_index=${e.index}`;
+        if (e.name) {
+          let titleWithEdition =
+            this.NFTName || this.$t('nft_details_page_title');
+          if (e.name) titleWithEdition += ` - ${e.name}`;
```
AuroraHuang22, line 788: `e.name` is already checked in the external conditions

```suggestion
       titleWithEdition += ` - ${e.name}`;
```

nwingt (approved):

AuroraHuang22 (approved):

PR #1896, opened 2024-10-16
[TAS-2320] 🚸 Add loading placeholder for <itemCard/>

AuroraHuang22:
截圖 2024-10-16 晚上9 51 09

### DIFF
--- a/src/components/NFTBook/ItemCard.vue
+++ b/src/components/NFTBook/ItemCard.vue
@@ -437,7 +437,6 @@ export default {
       this.isLoading = true;
       try {
         await this.lazyFetchNFTClassAggregatedData();
-      } catch (error) {
       } finally {
         this.isLoading = false;
       }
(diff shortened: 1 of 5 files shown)


src/components/NFTBook/ItemCard.vue
```diff
@@ -424,7 +434,13 @@ export default {
   },
   methods: {
     async fetchInfo() {
-      await this.lazyFetchNFTClassAggregatedData();
+      this.isLoading = true;
+      try {
+        await this.lazyFetchNFTClassAggregatedData();
+      } catch (error) {
```
williamchong, line 440: no need to `catch` here if we only need `finally`?

AuroraHuang22: [🔥 Remove unused catch](https://github.com/likecoin/liker-land/pull/1896/commits/42239182ecf3004272857a90f8b46ee9ff6286b8)

nwingt (approved):

PR #1903, opened 2024-10-19
🚸 Store nftId in url for reader page

williamchong:
mitigate issue where nftid fetch causes refresh to error out

### DIFF
--- a/src/pages/reader.vue
+++ b/src/pages/reader.vue
@@ -145,6 +145,15 @@ export default {
           );
           return;
         }
+        if (
+          (this.isLoginRequired || this.nftIsCustomMessageEnabled) &&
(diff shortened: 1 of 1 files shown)


src/pages/reader.vue
```diff
@@ -145,6 +145,13 @@ export default {
           );
           return;
         }
+        if (this.isLoginRequired || this.nftIsCustomMessageEnabled) {
+          if (this.nftId && !this.$route.query.nftId) {
```
nwingt, line 149: Combine conditions?

williamchong: Updated

nwingt (approved):

PR #1916, opened 2024-10-25
✨ Add tags in listing page

nwingt:
image


src/locales/zh-Hant.json
```diff
@@ -470,11 +471,12 @@
   "listing_page_header_sort_lower_price": "價錢低至高",
   "listing_page_header_sort_recommend": "推薦",
   "listing_page_header_total_books": "共 {num} 本書",
+  "listing_page_no_result": "沒有找到相關的書籍。",
```
williamchong, line 474:
```suggestion
  "listing_page_no_result": "沒有找到相關書籍。",

```

src/locales/zh-Hant.json
```diff
@@ -450,6 +450,7 @@
   "listing_page_cant_find_books": "找不到你的書籍？ ",
   "listing_page_cant_find_books_button": "告訴我們",
   "listing_page_cant_find_books_prefilled_message": "我想找",
+  "listing_page_clear_filter": "重設篩選",
```
williamchong, line 453:
```suggestion
  "listing_page_clear_filter": "清除篩選",

```

src/locales/zh-Hant.json
```diff
@@ -483,6 +485,7 @@
   "listing_page_select_language_title": "語言",
   "listing_page_select_language_zh": "繁體中文",
   "listing_page_sorter": "排序",
+  "listing_page_tag_not_found": "找不到這標籤",
```
williamchong, line 488:
```suggestion
  "listing_page_tag_not_found": "找不到此標籤",

```

src/server/api/util/airtable.js
```diff
@@ -68,6 +72,42 @@ async function fetchAirtableCMSItems(viewName, { pageSize = 100, offset }) {
...
+        offset,
+      },
+    }
+  );
+
+  const normalizedRecords = results.data.records.map(({ fields }) => {
+    const id = fields.ID;
+    const name = fields.Name;
+    const nameEn = fields['Name (Eng)'];
+    const isPublic = fields.Public;
```
williamchong, line 94: Should `isPublic` be filtered here instead of frontend?

src/store/modules/nft.js
```diff
@@ -985,6 +992,41 @@ const actions = {
       }))
     );
   },
+  async fetchBookstoreCMSProductsByTagId({ commit, getters }, tagId) {
+    try {
+      if (
+        !tagId ||
+        getters.nftGetBookstoreCMSProductsByTagIdHasFetched(tagId) ||
```
williamchong, line 999: not sure if we should name this `lazyFetchBookstoreCMSProductsByTagId` instead

src/pages/store/index.vue
```diff
@@ -960,6 +1100,17 @@ export default {
         1
       );
     },
+    handleClearFilterClick() {
+      this.$router.replace({
+        query: {
+          ...this.$route.query,
+          price: undefined,
+          drm_free: undefined,
+          lang: undefined,
```
williamchong, line 1109: When would these 3 be set?

williamchong (changes requested):

src/server/api/util/airtable.js
nwingt: `isPublic` means the tag is unlisted, the tag is not showing in the navigation, but can be accessed through the URL

PR #1917, opened 2024-10-25
⚡️ Simplify login logic and reduce await post login

williamchong:


src/mixins/wallet-login.js
```diff
@@ -54,15 +50,13 @@ export default {
...
-          ? this.initWallet(connection)
-          : this.initWalletAndLogin(connection));
+        const res = await this.initWalletAndLogin(connection);
 
         if (res) {
           logTrackerEvent(
             this,
             'user',
-            `connect_wallet_done${shouldSkipLogin ? '' : '_with_login'}`,
+            `connect_wallet_done$_with_login`,
```
nwingt, line 59:
```suggestion
            `connect_wallet_done_with_login`,
```

src/store/modules/wallet.js
```diff
@@ -869,15 +850,17 @@ const actions = {
   },
   async walletFetchSessionUserData(
     { dispatch },
-    { shouldSkipUserInfo = false } = {}
+    { awaitForWalletEvents = false } = {}
```
nwingt, line 853: where does this use?

nwingt (changes requested):

src/store/modules/wallet.js
williamchong: Is it not used yet, but setting this flag would mimic the original behaviour, which might be useful else where

nwingt (approved):

PR #1927, opened 2024-10-31
🔍️ Remove collector and add store tag in sitemap

williamchong:


src/config/sitemap.js
```diff
@@ -69,13 +69,16 @@ async function getSitemapRoutes() {
...
-  const collectors = ((collectorRes.data || {}).collectors || []).map(
-    c => c.account
-  );
-  const portfolioPageRoutes = [...new Set(creator.concat(collectors))].map(
-    id => `/${id}`
-  );
-  return nftDetailsPageRoutes.concat(portfolioPageRoutes).map(url => ({
+  const portfolioPageRoutes = creator.map(id => `/${id}`);
+
+  const tagIds = (cmsBookRes.data || {}.records || []).map(t => t.id);
```
nwingt, line 74:
```suggestion
  const tagIds = ((cmsBookRes.data || {}).records || []).map(t => t.id);
```

nwingt (changes requested):

PR #1961, opened 2024-11-17
🚑 Fix gifting

nwingt:


williamchong (approved):

AuroraHuang22 (approved):

src/pages/nft/class/_classId/index.vue
```diff
@@ -1471,8 +1471,8 @@ export default {
       await this.handleCollectFromEdition(this.giftSelectedValue, giftInfo);
       this.isGiftDialogOpen = false;
     },
-    handleGiftFromEditionSelector(selectedValue) {
-      this.giftSelectedValue = selectedValue;
+    handleGiftFromEditionSelector() {
```
nwingt, line 1474: @williamchong @AuroraHuang22  Should we also rename the func name & event name?

williamchong: yes go ahead

PR #1965, opened 2024-11-21
[TAS-1912] ✨ Store cart to user db

williamchong:


src/store/modules/cart.js
```diff
@@ -169,14 +172,29 @@ const actions = {
...
-  saveBookProductShoppingCart({ state }) {
+  async saveBookProductShoppingCart({ state, getters }) {
     saveShoppingCartToStorage(state.shoppingCartBookProductByIdMap, 'book');
+    if (getters.loginAddress) {
+      const cart = Object.values(state.shoppingCartBookProductByIdMap);
+      try {
+        if (cart.length) {
+          await this.$api.$post(postShoppingCart(), { cart });
+        } else {
+          await this.$api.$post(postShoppingCart(), { cart });
```
AuroraHuang22, line 183: why they are executing the same API ?
```suggestion
          await this.$api.$delete(deleteShoppingCart());
```

src/util/api/index.js
```diff
@@ -454,6 +454,10 @@ export const getUserV2Self = () => '/api/v2/users/self';
 export const postUserV2Login = () => '/api/v2/users/login';
 export const postUserV2Logout = () => '/api/v2/users/logout';
 
+export const getShoppingCart = () => '/api/v2/users/cart';
+export const postShoppingCart = () => '/api/v2/users/cart';
+export const deleteShoppingCart = () => '/api/v2/users/cart';
```
nwingt, line 459: How about?
```suggestion
export const postShoppingCart = getShoppingCart();
export const deleteShoppingCart = getShoppingCart();
```

nwingt (approved):

PR #1966, opened 2024-11-24
[TAS-2514] 💄Listing page header UI revamp

AuroraHuang22:
https://github.com/user-attachments/assets/db56f1ab-28f7-442f-b9d5-6a14629b531d

https://github.com/user-attachments/assets/dddf72c7-f75c-4b5c-9d42-8957d61d6c6a


williamchong (changes requested):
Need to confirm the arrows for tag list on mobile is clickable or swipable

williamchong (approved):

williamchong (changes requested):
image
Should also implement suggestion in index

image
If I press `Enter` here without typing anything, can we automatically choose the 1st suggested keyword and start the search?

nwingt (approved):

PR #1968, opened 2024-11-25
[TAS-2526] 🐛 Fix gifting edition selection

### LINKED ISSUE #1967: 贈送書本 ignoring edition selection
bug: when using 贈送書本, the system uses edition 0 no matter which one i pick.

also (a relatively minor ux issue), it doesn't inherit the edition chosen in the listing UI before i click 贈送書本

Screenshot 2024-11-25 at 17 53 25

nwingt:
Fix #1967
- Use computed prop for `selectedValue`(bad naming, should name `selectedEditionIndex`) instead of a separated component data to prevent data out-sync
- Sync `selectedValue` across the product page & the gift dialog by sharing the same prop

@AuroraHuang22 Notice that `update:value` is only available in Vue 3

### DIFF
--- a/src/components/NFTBook/GiftDialog.vue
+++ b/src/components/NFTBook/GiftDialog.vue
@@ -25,6 +25,7 @@
         :stock="item.stock"
         :is-selected="item.value === selectedValue"
         :current-price="item.price"
+        :is-disabled="item.isPhysicalOnly"
         :default-price="item.defaultPrice"
         preset="dialog"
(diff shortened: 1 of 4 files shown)


williamchong (approved):

AuroraHuang22 (approved):

nwingt:
Disable gifting for physical books

…]()

AuroraHuang22 (approved):

williamchong (approved):

PR #1976, opened 2024-12-05
[TAS-2509] 👽️ Disable auto font sizing on mobile

williamchong:


src/pages/reader/epub.vue
```diff
@@ -368,6 +368,12 @@ export default {
           ),
         });
         const cfi = this.resumeFromLocalStorage();
+        this.rendition.themes.default({
+          p: {
```
nwingt, line 372:
```suggestion
          body: {
```

nwingt (changes requested):

PR #1980, opened 2024-12-06
[TAS-2509] 💩 Force EPUB direction to ltr

nwingt:


src/pages/reader/epub.vue
```diff
@@ -372,6 +372,7 @@ export default {
           body: {
             '-webkit-text-size-adjust': 'none',
             'text-size-adjust': 'none',
+            direction: 'ltr',
```
williamchong, line 375:
```suggestion
            direction: 'ltr', // mitigate epubjs mixing up dir & page-progression-direction
```

PR #2006, opened 2024-12-30
[TAS-2919] 🔥 Disable Christmas coupon functionality

AuroraHuang22:
Deploy after 3-2-1


src/pages/nft/class/_classId/index.vue
```diff
@@ -444,7 +444,6 @@ import crossSellMixin from '~/mixins/cross-sell';
 import navigationListenerMixin from '~/mixins/navigation-listener';
 import utmMixin from '~/mixins/utm';
 import alertMixin from '~/mixins/alert';
-import couponMixin from '~/mixins/coupon';
```
williamchong, line 447: Maybe we can keep all the coupon mixins, but make sure `getApplicableCoupon` return nothing for now?

williamchong (changes requested):

src/pages/nft/class/_classId/index.vue
AuroraHuang22: [🔥 Disable christmas coupon functionality](https://github.com/likecoin/liker-land/pull/2006/commits/e375bec7f6658092a538e7be58a96c1c1931d71f)

williamchong (approved):

PR #2007, opened 2024-12-31
[TAS-2787] 📈 Add logging for PDF download event

AuroraHuang22:
listen event from https://github.com/likecoin/pdf.js/pull/10

### DIFF
--- a/src/pages/reader/pdf.vue
+++ b/src/pages/reader/pdf.vue
@@ -82,8 +82,7 @@ export default {
           );
           this.fileData = null;
         }
-      }
-      if (event.data.event === 'pdfDownload') {
+      } else if (event.data.event === 'pdfDownload') {
         logTrackerEvent(
           this,
           'ReaderPdf',
--- a/src/pages/reader/pdf.vue
+++ b/src/pages/reader/pdf.vue
@@ -8,7 +8,7 @@
         :type="progressIndicatorType"
         :value="progressPercent"
       />
-      <span class="text-white text-center">{{ progressLabelText }}</span>
+      <span class="text-center text-white">{{ progressLabelText }}</span>
     </div>
     <iframe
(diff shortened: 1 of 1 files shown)


src/pages/reader/pdf.vue
```diff
@@ -83,6 +83,19 @@ export default {
...
+          this,
+          'ReaderPdf',
+          'PDFDownloaded',
+          {
+            classId: this.classId,
+            filename: event.data.filename,
+          },
+          1
+        );
+        console.log(`PDF downloaded: ${event.data.filename}`);
```
williamchong, line 97: Please remove unneeded `console.log`

src/pages/reader/pdf.vue
```diff
@@ -83,6 +83,19 @@ export default {
...
+        logTrackerEvent(
+          this,
+          'ReaderPdf',
+          'PDFDownloaded',
+          {
+            classId: this.classId,
+            filename: event.data.filename,
+          },
+          1
+        );
```
williamchong, line 96: 4th param does not expect object
and maybe we should name it as present tense since download is not completed at this point

```suggestion
        logTrackerEvent(
          this,
          'ReaderPdf',
          'ReaderDownloadPdf',
          classId: this.classId,
          1
        );
```

williamchong (changes requested):

src/pages/reader/pdf.vue
```diff
@@ -83,6 +83,15 @@ export default {
           this.fileData = null;
         }
       }
+      if (event.data.event === 'pdfDownload') {
```
williamchong, line 86:
```suggestion
      } else if (event.data.event === 'pdfDownload') {
```

williamchong (changes requested):

PR #2011, opened 2025-01-09
[TAS-3007] ➕ Integrate Poptin

nwingt:


src/nuxt.config.js
```diff
@@ -147,6 +147,11 @@ const nuxtConfig = {
         src: 'https://js.stripe.com/v3',
         async: true,
       },
+      {
+        hid: 'pixel-script-poptin',
+        src: 'https://cdn.popt.in/pixel.js?id=771c46b9e377e',
+        async: true,
+      },
```
williamchong, line 154: Add preload if this is global?

nwingt: Added

PR #2015, opened 2025-01-15
[TAS-2718]💄 Enhance book info section by adding additional details

AuroraHuang22:
截圖 2025-01-16 上午11 03 48
截圖 2025-01-16 上午11 04 13


src/store/modules/nft.js
```diff
@@ -879,6 +885,8 @@ const actions = {
       signedMessageText: data.signedMessageText,
       enableSignatureImage: data.enableSignatureImage,
       isHidden: data.isHidden,
+      tableOfContents: data.tableOfContents,
+      blurb: data.blurb,
```
williamchong, line 889: Please don't add this yet, we would probably be using `https://schema.org/Review` in either iscn or listing data, but it is not there in book press and not confirmed yet

williamchong (changes requested):

williamchong (approved):

src/components/NFTBook/ItemCard.vue
```diff
@@ -152,10 +152,63 @@
...
+                :content="iscnWorkAuthorDescription"
+              />
+              <NFTBookContentBlock
+                v-if="nftTableContent"
+                :title="$t('nft_details_page_label_table_content')"
+                :content="nftTableContent"
+              />
+              <div
+                v-if="!isExpanded"
+                class="absolute bottom-0 w-full h-[80px] bg-gradient-to-b from-transparent to-white"
```
nwingt, line 195: Should this add `pointer-events-none`?

src/components/NFTBook/ItemCard.vue
```diff
@@ -152,10 +152,63 @@
...
+                'h-full',
+                { 'max-h-[240px]': !isExpanded },
+                'overflow-y-hidden',
+                'flex flex-col',
+                'gap-[40px]',
+              ]"
+            >
+              <p
+                :class="[
+                  'blick',
```
nwingt, line 175: what is `blick` class?

AuroraHuang22: block la~
[💄 Update styles](https://github.com/likecoin/liker-land/pull/2015/commits/e93bde4bbd1b94b17086939d31519eb6dc70601c)

nwingt (approved):

PR #2018, opened 2025-01-16
[TAS-3080] 🚸 Enhance search behavior

AuroraHuang22:
https://github.com/user-attachments/assets/208a677c-b3c8-4965-9264-eeae487ffdb6

### DIFF
--- a/src/components/SearchBar.vue
+++ b/src/components/SearchBar.vue
@@ -39,7 +39,6 @@
         'bg-shade-gray',
         'gap-[12px]',
         'h-full',
-        'group',
         isSearchOpen && 'pl-[16px] pr-[4px]',
         'rounded-[0px] desktop:rounded-[10px]',
         'transition-[width]',
@@ -105,7 +104,7 @@ export default {
     searchButtonProps() {
       if (!this.isSearchOpen) return null;
       switch (this.searchActionState) {
-        case 'typing':
+        case SEARCH_ACTIONS.TYPING:
           return {
             size: 'small',
             preset: 'tertiary',
@@ -114,8 +113,8 @@ export default {
             text: this.$t('listing_page_search'),
             onClick: this.handleSubmit,
           };
-        case 'submitted':
-        case 'idle':
+        case SEARCH_ACTIONS.SUBMITTED:
+        case SEARCH_ACTIONS.IDLE:
           return {
             size: 'tiny',
             preset: 'plain',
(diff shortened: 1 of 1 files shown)


src/components/SearchBar.vue
```diff
@@ -38,40 +38,50 @@
...
         'h-full',
+        isSearchOpen && 'pl-[16px] pr-[4px]',
         'rounded-[0px] desktop:rounded-[10px]',
         'transition-[width]',
-        'duration-[5000]',
+        'duration-[300]',
         'ease-in-out',
       ]"
     >
       <IconSearch />
```
williamchong, line 49: Can we make this icon clickable so that it also serve as the `open`/`submit` button if people can't press enter?

src/components/SearchBar.vue
```diff
@@ -123,13 +163,24 @@ export default {
       }
       return this.$t('gutenberg_search_placeholder');
     },
-    handleEnter() {
-      if (!this.searchQuery || this.searchQuery.trim() === '') {
-        const fallbackKeyword =
-          this.randomKeywords?.[0] || this.$t('gutenberg_search_placeholder');
-        this.$emit('input', fallbackKeyword);
+    handleInput(event) {
+      this.currentSearchQuery = event.target.value;
```
williamchong, line 167: I think we don't need to set `currentSearchQuery` explicitly in `@input` if it is in `v-model`

williamchong (changes requested):

src/components/SearchBar.vue
AuroraHuang22: we have submit button when open
截圖 2025-01-22 下午6 47 29

williamchong (approved):

src/components/SearchBar.vue
```diff
@@ -38,40 +38,50 @@
         'items-center',
         'bg-shade-gray',
         'gap-[12px]',
-        isSearchOpen && 'pl-[16px] pr-[4px]',
         'h-full',
+        isSearchOpen && 'pl-[16px] pr-[4px]',
```
nwingt, line 42:
```suggestion
       { 'pl-[16px] pr-[4px]': isSearchOpen },
```

nwingt (changes requested):

src/components/SearchBar.vue
AuroraHuang22: [♻️ Refactor SearchBar styles](https://github.com/likecoin/liker-land/pull/2018/commits/e32ad68305c80bf5dbf82c90225e0e23a2ac7e3d)

nwingt (approved):

PR #2019, opened 2025-01-16
[TAS-3047] 👔 Allow setting cross sell probability for each item

williamchong:


nwingt (approved):

src/mixins/cross-sell.js
```diff
@@ -55,13 +55,16 @@ export default {
       );
     },
     isCrossSellEnabled() {
+      const crossSellInfo = !CROSS_SELL_WHITE_LIST.find(item => {
+        const id = item?.id || item;
+        return id === this.classId || id === this.collectionId;
+      });
```
nwingt, line 61:
```suggestion
      const crossSellInfo = CROSS_SELL_WHITE_LIST.find(item => {
        const id = item?.id || item;
        return id === this.classId || id === this.collectionId;
      });
```

nwingt (changes requested):

PR #2020, opened 2025-01-17
[TAS-3068] 💩 Handle EPUB with pre-paginated layout & no spread

nwingt:


src/pages/reader/epub.vue
```diff
@@ -368,13 +368,21 @@ export default {
...
-          body: {
-            '-webkit-text-size-adjust': 'none',
-            'text-size-adjust': 'none',
-            direction: 'ltr', // mitigate epubjs mixing up dir & page-progression-direction
-          },
-        });
+
+        const metadata = await this.book.loaded.metadata;
+        const bodyCSS = {
+          color: '#333',
```
williamchong, line 374: Why add #333 though

nwingt: liker.land default text color in bookstore
https://github.com/likecoin/liker-land/blob/23ff92e24d95647d42a719201aabc7926ebe026a/src/components/NFTBook/ItemCardV2.vue#L25

PR #2025, opened 2025-01-21
✨ Add evm migration API

williamchong:


nwingt (approved):

src/server/api/routes/users/v2/wallet.js
```diff
@@ -58,4 +62,82 @@ router.get('/wallet', async (req, res, next) => {
   }
 });
 
+router.post('/wallet/evm/migrate', async (req, res, next) => {
+  try {
+    const key = req.headers['x-likerland-api-key'];
+    const { wallet } = req.query;
+    const { evmWallet } = req.body;
```
nwingt, line 69: why not all put in body?

williamchong: Trying to keep consistent that `const { wallet } = req.query;` is the resource being targeted, `evmWallet` is part of the payload for POST

PR #2026, opened 2025-01-21
[TAS-2972] 🎨 Clean up bookstore logic for tags & sorting

nwingt:


src/constant/store.js
```diff
@@ -1,5 +1,4 @@
 export const SORTING_OPTIONS = {
-  RECOMMEND: 'recommend',
```
nwingt, line 2: `recommend` is replaced by `featured` tag

src/pages/store/featured.vue
```diff
@@ -1,12 +1,10 @@
 <script>
-import { SORTING_OPTIONS } from '~/constant/store';
-
 export default {
   fetch({ redirect, localeLocation, query }) {
     redirect(
       localeLocation({
         name: 'store',
-        query: { ...query, sort: SORTING_OPTIONS.RECOMMEND },
+        query: { ...query },
```
nwingt, line 7: Default sorting is controlled by store page

williamchong, line 7:
```suggestion
        query,
```

williamchong (changes requested):

williamchong (approved):

PR #2027, opened 2025-01-22
🚨 Fix route missing id

nwingt:
Screenshot 2025-01-22 at 09 36 19


williamchong (approved):
How can a card not have product owner though?

nwingt:
> How can a card not have product owner though?

before data is fetched

PR #2028, opened 2025-01-22
✨ Clear CMS cache by query string

nwingt:


src/pages/index.vue
```diff
@@ -903,14 +903,15 @@ export default {
       searchQuery: '',
     };
   },
-  async fetch({ store }) {
+  async fetch({ store, route }) {
+    const { t } = route.query;
```
williamchong, line 907: What would be the initial value of t and how would it be defined?

nwingt: assume `t` is unix timestamp

PR #2029, opened 2025-01-22
[TAS-1178] ✨ Support pagination for bookstore

nwingt:
Also replaced latest bookstore items with CMS to display collection

image


src/store/modules/nft.js
```diff
@@ -44,7 +44,9 @@ const state = () => ({
   trendingNFTClassIdList: [],
   bookstoreLatestItems: [],
   bookstoreCMSProductsByTagIdMap: {},
+  bookstoreCMSProductsPaginationOffsetByTagIdMap: {},
   bookstoreCMSProductsByTagIdIsFetchingMap: {},
+  bookstoreCMSProductsByTagIdIsFetchingMoreMap: {},
```
williamchong, line 49: These looks like they could be in a same map, with something like
```
{
  products: [],
  offset,
  isFetching,
  isFetchingMore,
}
```

src/store/modules/nft.js
```diff
@@ -1006,42 +1028,56 @@ const actions = {
...
-        value: false,
-      });
+      commit(fetchingMutation, { id: tagId, value: false });
     }
   },
+  async fetchMoreBookstoreCMSProductsByTagId({ dispatch, getters }, { tagId }) {
+    const offset = getters.nftGetBookstoreCMSProductsPaginationOffsetByTagId(
+      tagId
+    );
+    if (!offset) return;
```
williamchong, line 1073: Not sure if we should just let it fetch with `offset = 0` if `nftGetBookstoreCMSProductsByTagIdHasFetched` is not set in this case

williamchong (approved):

src/store/modules/nft.js
nwingt: but `offset` is not a number
image

williamchong: Is the guard trying to guard the case where we have not set offset yet, or the list has ended so there is no offset from API? If latter then no problem

PR #2045, opened 2025-02-04
🔧 Get 100 items for listing tags

williamchong:


nwingt (commented):
how about just changing the default here
https://github.com/likecoin/liker-land/blob/7c2e76eccf6c1a7b13a05330cd00356efd058a9d/src/util/api/index.js#L615

williamchong:
> how about just changing the default here
> 
> https://github.com/likecoin/liker-land/blob/7c2e76eccf6c1a7b13a05330cd00356efd058a9d/src/util/api/index.js#L615

i don't know, what was the reason behind setting 30 in the pagination PR?
If it is just the landing needs 100, other tags could keep using 30

PR #2050, opened 2025-02-11
[TAS-3208] 💄 Add home/back button in PDF reader

nwingt:
image

### DIFF
--- a/src/pages/reader/pdf.vue
+++ b/src/pages/reader/pdf.vue
@@ -1,8 +1,53 @@
 <template>
-  <div class="w-full h-full">
+  <div class="fixed inset-0 flex flex-col">
+    <header class="border-b border-[#ccc]">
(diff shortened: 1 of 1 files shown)


src/pages/reader/pdf.vue
```diff
@@ -1,33 +1,72 @@
...
+          <NuxtLink
+            :to="
+              localeLocation(
+                getAddress
+                  ? {
+                      name: 'id',
+                      params: { id: getAddress },
+                      query: { tab: 'collected' },
+                    }
+                  : { name: 'index' }
```
williamchong, line 15: to class page maybe?

nwingt: should change epub reader too?

williamchong: yes, merge this first if you want to do it in new pr

PR #2054, opened 2025-02-13
🚸 Update reader back button behavior

nwingt:


src/pages/reader/pdf.vue
```diff
@@ -5,17 +5,13 @@
...
           <NuxtLink
             :to="
-              localeLocation(
-                getAddress
-                  ? {
-                      name: 'id',
-                      params: { id: getAddress },
-                      query: { tab: 'collected' },
-                    }
-                  : { name: 'index' }
```
williamchong, line 15: How about change the check to `getSessionWallet`, if exists go to `bookshelf`, if not go to `classId` page?

williamchong (changes requested):

src/pages/reader/pdf.vue
nwingt: Updated

williamchong (approved):

PR #2055, opened 2025-02-13
[TAS-3235] 🐛 Fix unable to read not login required book when logged in

nwingt:
Cannot read book when logged in if no login is required to read book

### DIFF
--- a/src/pages/reader.vue
+++ b/src/pages/reader.vue
@@ -109,9 +109,10 @@ export default {
           this.lazyFetchNFTOwners(),
         ]);
         if (
-          (!this.userCollectedCount && this.isLoginRequired) ||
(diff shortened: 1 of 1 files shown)


williamchong (commented):
I think we need some clarification on behaviour

If an NFT ID of someone else's book is set, and user is either guest or logged in, are they expected to be able to read other ppl's specific version of book?
Should we just bounce it back to a version without NFT ID?

nwingt:
I think bounce it back to a version without NFT ID is more appropriate

williamchong:
> I think bounce it back to a version without NFT ID is more appropriate

Then we might want to handle `!!nftId && this.this.nftCollectorWalletAddress && this.nftCollectorWalletAddress !== this.getSessionWallet` case differently

PR #2067, opened 2025-02-20
[TAS-3267] 💬 Update cross-sell dialog title for Bottle Shiu

nwingt:
image


williamchong (approved):
Could become generalized using `author` field if this works

PR #2082, opened 2025-02-25
♻️ Refactor nft vs NFT vs iscn data parsing in mixin

williamchong:
Base on https://github.com/likecoin/liker-land/pull/2081

### DIFF
--- a/src/mixins/nft.js
+++ b/src/mixins/nft.js
@@ -1,7 +1,6 @@
 import { mapActions, mapGetters } from 'vuex';
 
 import {
-  APP_LIKE_CO_VIEW,
   LIKECOIN_CHAIN_API,
   TX_STATUS,
   LIKECOIN_NFT_API_WALLET,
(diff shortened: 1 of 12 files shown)


src/mixins/nft.js
```diff
@@ -113,21 +110,37 @@ export default {
       'uiTxNFTStatus',
     ]),
     NFTClassMetadata() {
```
nwingt, line 112: why not changing this?

src/pages/nft/class/_classId/index.vue
```diff
@@ -716,10 +716,10 @@ export default {
           name: 'Writing NFT',
         },
         sku: this.classId,
-        isbn: this.iscnData?.contentMetadata?.isbn,
+        isbn: this.contentMetadata.isbn,
         datePublished:
-          this.iscnData?.contentMetadata?.datePublished ||
-          this.iscnData?.recordTimestamp,
+          this.contentMetadata.datePublished ||
```
nwingt, line 721:
```suggestion
          this.contentMetadata?.datePublished ||
```

src/mixins/nft.js
williamchong: `contentMetadata` is combined of `NFTClassMetadata`, `nftClassMetadata` and `iscnData`, so these 3 have to be unchanged

src/pages/nft/class/_classId/index.vue
williamchong: `contentMetadata` should at least be an empty object, so the line below should be changed instead `this.contentMetadata.recordTimestamp`

williamchong: updated

nwingt (approved):

PR #2112, opened 2025-03-18
[TAS-3480] 💄 Fix broken NFT collection page & unify style

nwingt:
Before
image
image

After
localhost_3000_zh-Hant_nft_collection_col_book_931bb5a0-39e1-472e-963c-d542c9902ab4_ll_medium=listing

### DIFF
--- a/src/components/NFTBook/SpecTableItemDateReleased.vue
+++ b/src/components/NFTBook/SpecTableItemDateReleased.vue
@@ -12,7 +12,7 @@ export default {
       default: 'li',
     },
     date: {
-      type: String,
+      type: [String, Number],
       default: '',
     },
     isPublishedDate: {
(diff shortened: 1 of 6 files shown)


src/components/NFTBook/SpecTableItemDateReleased.vue
```diff
@@ -12,7 +12,7 @@ export default {
       default: 'li',
     },
     date: {
-      type: String,
+      type: [String, Number],
```
williamchong, line 15: In what case date would be Number?

nwingt: timestamp from the [API](https://api.like.co/likernft/collection/col_book_931bb5a0-39e1-472e-963c-d542c9902ab4)

PR #2151, opened 2025-05-19
[TAS-3611] 👽 Add chain filter for Airtable API

nwingt:


src/server/api/util/airtable.js
```diff
@@ -33,7 +38,7 @@ async function fetchAirtableCMSProductsByTagId(
         Authorization: `Bearer ${AIRTABLE_API_SECRET}`,
       },
       params: {
-        filterByFormula: 'NOT(Hidden)',
+        filterByFormula: `AND(NOT(Hidden), Chain = "${getChainId()}")`,
```
williamchong: Maybe we should allow override via qs

nwingt: Updated

williamchong (approved):

PR #2154, opened 2025-06-09
🐛 Fix evm book redirect to 3ook

williamchong:


src/pages/nft/class/_classId/index.vue
```diff
@@ -656,7 +656,7 @@ export default {
         });
       }
     }
-    const priceInfo = store.getters.getNFTBookStorePricesByClassId(classId);
+    const priceInfo = store.getters.getNFTBookStoreInfoByClassId(classId);
```
nwingt: better change `priceInfo` to `storeInfo` for clarity

PR #2156, opened 2025-06-16
🎨 Update migration links to use UTM parameters

AuroraHuang22:


src/constant/index.js
```diff
@@ -530,4 +530,4 @@ export const MIGRATION_URL = IS_TESTNET
   ? 'https://likenft-migration.pandawork.com/'
   : 'https://v3-book-migration.like.co/';
 
-export const IS_CLAIM_DISABLED = true;
+export const IS_MIGRATING = true;
```
nwingt:
```suggestion
export const IS_MIGRATION_STARTED = true;
```

AuroraHuang22: [🚸 Show migration popup if is migration](https://github.com/likecoin/liker-land/pull/2156/commits/da314eb79a317a981d53fd678eaf9dc5f005891f)

src/layouts/default.vue
```diff
@@ -232,8 +233,11 @@ export default {
   methods: {
     ...mapActions(['uiCloseTxModal']),
     onClickAlertBanner() {
+      const URL = getBookComUrlWithUTM({
```
nwingt: URL is a class
```suggestion
      const url = getBookComUrlWithUTM({
```

src/layouts/default.vue
```diff
@@ -232,8 +233,11 @@ export default {
...
+        medium: 'banner',
+      });
       logTrackerEvent(this, 'alert_banner', 'alert_banner_click_', '', 1);
-      window.open(BOOK_COM_DOMAIN, '_blank');
+      window.open(URL, '_blank');
```
nwingt:
```suggestion
      window.open(url, '_blank');
```

nwingt (changes requested):

williamchong (approved):

